PR #2, opened 2019-07-20
Added analyzer that verify if method provider for Setup() is overridable

cezarypiatek:
This implementation handles only `Setup()` for methods. The case with properties is currently ignored because is much more complex, especially that `Moq` supports recursive mocking for properties. I've set the default severity level for `ERROR` because it always fails in runtime.


Litee:
Thanks! I have upgraded the master to work with VS 2019. Do you mind re-basing and updating this PR?

cezarypiatek:
Thanks for responding. I've just rebased my branch (and also fixed an issue with setup for properties)

PR #3, opened 2019-07-28
New feature: Mock.As() should take interfaces only

cezarypiatek:
I've added an analyzer that verifies if the type parameter accepted by Moq.As represents an interface.


cezarypiatek:
@Litee would you mind to review and merge my changes?

Litee:
Sorry for the delay, merged now and released under 0.0.8 version.

Issue #8, opened 2022-07-02
NuGet Package does not show up as an analyzer in VS Solution Explorer


BrightLight:
The package that is available on nuget.org (and here on GitHub) does not show up in the Visual Studio Solution explorer under "Analyzers". I think the reason for this is that the package stores the dll in `\lib\netstandard1.3` instead of `\analyzers\dotnet\cs`
![NuGetPackageExplorer_Moq Analyzers_0 0 9](https://user-images.githubusercontent.com/1859278/176994537-b0acaa4a-b13e-4d76-a761-46ba2372d64d.png)

I tried to investigate this a bit, but both the `Diagnostic.nuspec` and `Moq.Analyzers.csproj` look fine. Both use "analyzers\dotnet\cs" as the path. I don't know which of these two is used to built the actual nuget, my guess is `Diagnostic.nuspec` as it's using 0.0.9 while the csproj is still on 0.0.6.

Could you please upload an updated nuget with the correct path?

rjmurillo:
Hi @BrightLight. Thanks for the issue report. We are working toward a new release that will correct this issue.

rjmurillo:
Hi @BrightLight this has been fixed in version [0.1.0](https://www.nuget.org/packages/Moq.Analyzers/0.1.0#readme-body-tab).

PR #30, opened 2024-05-31
Add .nupkg baseline with Verify

MattKotsenas:
Add a test that inspects the `.nupkg` created as part of the build (via `<GeneratePackageOnBuild>`) and dumps the manifest and file structure to prevent accidental regressions in the package format.

### DIFF
--- a/Source/Moq.Analyzers.Test/PackageTests.Baseline#contents.verified.txt
+++ b/Source/Moq.Analyzers.Test/PackageTests.Baseline#contents.verified.txt
@@ -1,0 +1,12 @@
+﻿/
+|-- Moq.Analyzers.nuspec
+|-- analyzers
+|   |-- dotnet
+|   |   |-- cs
+|   |   |   |-- Moq.Analyzers.dll
+|-- lib
+|   |-- netstandard2.0
+|   |   |-- Moq.Analyzers.dll
+|-- tools
+|   |-- install.ps1
+|   |-- uninstall.ps1
--- a/Source/Moq.Analyzers.Test/PackageTests.Baseline#manifest.verified.nuspec
+++ b/Source/Moq.Analyzers.Test/PackageTests.Baseline#manifest.verified.nuspec
@@ -1,0 +1,17 @@
+﻿<package>
+  <metadata>
+    <id>Moq.Analyzers</id>
+    <version>********</version>
+    <authors>Andrey "Litee" Lipatkin</authors>
+    <licenseUrl>https://github.com/Litee/moq.analyzers/blob/master/LICENSE</licenseUrl>
+    <projectUrl>https://github.com/Litee/moq.analyzers</projectUrl>
+    <description>Roslyn analyzer that helps to write unit tests using Moq mocking library by highlighting typical errors and suggesting quick fixes. Port of Resharper extension to Roslyn. Find the full list of detected issues at project GitHub page.</description>
+    <releaseNotes>Upgraded to support Visual Studio 2019</releaseNotes>
+    <copyright>2015-2019 Andrey Lipatkin</copyright>
+    <tags>moq, mock, test, analyzers</tags>
+    <repository type="git" commit="****************************************" />
+    <dependencies>
+      <group targetFramework=".NETStandard2.0" />
+    </dependencies>
+  </metadata>
+</package>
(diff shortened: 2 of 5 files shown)


Source/Moq.Analyzers.Test/PackageTests.Baseline#contents.verified.txt
```diff
@@ -0,0 +1,12 @@
+﻿/
+|-- Moq.Analyzers.nuspec
+|-- analyzers
+|   |-- dotnet
+|   |   |-- cs
+|   |   |   |-- Moq.Analyzers.dll
+|-- lib
```
MattKotsenas, line 7: Since this is an analyzer package and there's no "shipping" code, I don't think the `lib/` directory should be in the package. Tracked by #31.

Source/Moq.Analyzers.Test/PackageTests.Baseline#manifest.verified.nuspec
```diff
@@ -0,0 +1,17 @@
...
+    <authors>Andrey "Litee" Lipatkin</authors>
+    <licenseUrl>https://github.com/Litee/moq.analyzers/blob/master/LICENSE</licenseUrl>
+    <projectUrl>https://github.com/Litee/moq.analyzers</projectUrl>
+    <description>Roslyn analyzer that helps to write unit tests using Moq mocking library by highlighting typical errors and suggesting quick fixes. Port of Resharper extension to Roslyn. Find the full list of detected issues at project GitHub page.</description>
+    <releaseNotes>Upgraded to support Visual Studio 2019</releaseNotes>
+    <copyright>2015-2019 Andrey Lipatkin</copyright>
+    <tags>moq, mock, test, analyzers</tags>
+    <repository type="git" commit="****************************************" />
+    <dependencies>
+      <group targetFramework=".NETStandard2.0" />
```
MattKotsenas, line 14: Since this is an analyzer only, there's no shipping code, and hence no dependencies, so I don't think this should be here. Tracked by #31.

Source/Moq.Analyzers.Test/PackageTests.Baseline#manifest.verified.nuspec
```diff
@@ -0,0 +1,17 @@
+﻿<package>
+  <metadata>
+    <id>Moq.Analyzers</id>
+    <version>********</version>
+    <authors>Andrey "Litee" Lipatkin</authors>
+    <licenseUrl>https://github.com/Litee/moq.analyzers/blob/master/LICENSE</licenseUrl>
```
MattKotsenas, line 6: We should switch from the old `<licenseUrl>` to the new `<licenseExpression>`. Tracked by #31.

Source/Moq.Analyzers.Test/PackageTests.Baseline#manifest.verified.nuspec
```diff
@@ -0,0 +1,17 @@
...
+    <projectUrl>https://github.com/Litee/moq.analyzers</projectUrl>
+    <description>Roslyn analyzer that helps to write unit tests using Moq mocking library by highlighting typical errors and suggesting quick fixes. Port of Resharper extension to Roslyn. Find the full list of detected issues at project GitHub page.</description>
+    <releaseNotes>Upgraded to support Visual Studio 2019</releaseNotes>
+    <copyright>2015-2019 Andrey Lipatkin</copyright>
+    <tags>moq, mock, test, analyzers</tags>
+    <repository type="git" commit="****************************************" />
+    <dependencies>
```
MattKotsenas, line 13: We probably want to mark ourselves as a dev dependency. Tracked in #31.

Source/Moq.Analyzers.Test/PackageTests.Baseline#contents.verified.txt
rjmurillo: Related to #8?

MattKotsenas: Yes, likely. `0.0.8` puts it in both places like this change has. `0.0.9` is missing the analyzer.

So I think #8 will be fixed in the next release either way, and also we should ship the analyzer as a lib.

rjmurillo (approved):

PR #52, opened 2024-06-04
Add cancellation tokens to methods that have a parameter but we don't pass

### LINKED ISSUE #43: Enable analyzer to check for cancellation token method overloads
- [ ] Turn on analyzer
- [ ] Baseline existing code
- [ ] Create backlog for fixes

rjmurillo:
Most often this was while getting the type information from the semantic model. In cases where the asynchronous operation is cancelled, the underlying calls would not be aborted because we did not propagate forward the cancellation token. This was not picked up by analyzer CA2016 because the cancellation token is not provided on the method used by `DiagnosticAnalyzer`, but rather on the `SyntaxNodeAnalysisContext` instance passed in. 

The issue was corrected by adding a new analyzer package and enabling an analyzer rule within that package. Both rules are set to error in `.editorconfig` to avoid introducing new code with this issue

Fixes #43

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -375,3 +375,7 @@ dotnet_naming_style.s_camelcase.required_suffix =
 dotnet_naming_style.s_camelcase.word_separator =
 dotnet_naming_style.s_camelcase.capitalization = camel_case
 
+# MA0040: Forward the CancellationToken parameter to methods that take one
+dotnet_diagnostic.MA0040.severity = error
+# Async analyzer
+dotnet_diagnostic.CA2016.severity = error
--- a/.editorconfig
+++ b/.editorconfig
@@ -184,9 +184,6 @@ csharp_space_between_square_brackets = false
 csharp_preserve_single_line_blocks = true
 csharp_preserve_single_line_statements = true
 
-# Async analyzer
-dotnet_diagnostic.CA2016.severity = error
-
 #### Naming styles ####
 [*.{cs,vb}]
 
--- a/Source/Moq.Analyzers/Moq.Analyzers.csproj
+++ b/Source/Moq.Analyzers/Moq.Analyzers.csproj
@@ -19,10 +19,14 @@
     <PackageReleaseNotes>A changelog is available at https://github.com/Litee/moq.analyzers/releases</PackageReleaseNotes>
     <Copyright>2017 Andrey Lipatkin</Copyright>
     <PackageTags>moq, tdd, mocking, mocks, unittesting, agile, unittest, mock, test, analyzers</PackageTags>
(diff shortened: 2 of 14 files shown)


Source/Moq.Analyzers/Moq.Analyzers.csproj
```diff
@@ -19,10 +19,14 @@
     <PackageReleaseNotes>A changelog is available at https://github.com/Litee/moq.analyzers/releases</PackageReleaseNotes>
     <Copyright>2017 Andrey Lipatkin</Copyright>
     <PackageTags>moq, tdd, mocking, mocks, unittesting, agile, unittest, mock, test, analyzers</PackageTags>
-    <DevelopmentDependency>true</DevelopmentDependency> 
+    <DevelopmentDependency>true</DevelopmentDependency>
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Meziantou.Analyzer">
```
MattKotsenas, line 26: We should lift this and the `Roslynator.Analyzers` out of this .csproj and into `//build/targets/compiler/Compiler.props` (or create another vertical slice for code analysis, your call).

MattKotsenas (approved):
Left a single comment about lifting code analysis packages into a build slice. Also happy to do that as a follow up change (if so, file an issue).

Thanks!

Source/Moq.Analyzers/Moq.Analyzers.csproj
rjmurillo: I moved it and fixed issues that were in the test code. I can't properly fix the issue in test until xUnit v3. See [xunit/#345](https://github.com/xunit/xunit/issues/345)

.editorconfig
```diff
@@ -184,6 +184,9 @@ csharp_space_between_square_brackets = false
 csharp_preserve_single_line_blocks = true
 csharp_preserve_single_line_statements = true
 
+# Async analyzer
+dotnet_diagnostic.CA2016.severity = error
```
MattKotsenas, line 188: Duplicated below. Can we remove and have a single definition?

MattKotsenas (changes requested):

.editorconfig
rjmurillo: This should be the new syntax? `true:error`

MattKotsenas: Let me check. There's the ID ones XXDDDD and there are the named ones...

MattKotsenas: For an ID one or appears to just be the severity: https://github.com/dotnet/roslyn/blob/59644a32be5ad010d87065e8c8238d21f9f6eb5e/.editorconfig#L238

PR #57, opened 2024-06-05
Refactor tests to simplify and inline test cases

MattKotsenas:
Mechanical refactor as a prerequisite for improving test coverage and moving from a custom test harness over to the Microsoft.CodeAnalysis.Testing harness.

Each test class used the same procedure:

1. Split the `Data\*.cs` file into multiple files, one per test method
2. Update the test harness split into multiple `[Fact]`s, one per `Data` class and accept baselines
3. Remove unused code from the `Data` class and accept baselines
4. Inline the `Data` class into the test harness

The CodeFix test is a bit different. Because the diagnostics and code fix should be tested together, I inlined those `Data` classes into a string. A future refactor to use the Microsoft.CodeAnalysis.Testing harness will further clean those up.

Added a few TODOs where test coverage appears unnecessary or missing, but avoided adding it in this change due to the difficulty of review.


rjmurillo (approved):
Thanks @MattKotsenas

PR #60, opened 2024-06-06
Extract code analysis to its own build slice and update analyzers to latest versions

MattKotsenas:
This is a preparatory refactoring change for the test updates.

1. Move code analysis build settings from the `compiler` slice to its own `codeanalysis` slice
2. Update `Microsoft.CodeAnalysis.Analyzers`, `Roslynator.Analyzers`, and `StyleCop.Analyzers` to their latest versions
3. Fix some duplicate definitions between the slices and the projects
4. Move `Microsoft.CodeAnalysis.CSharp.Workspaces` out of the code analysis slice because it's part of the analyzer's public API and not a private code analyzer


rjmurillo (approved):
LGTM. Thanks for organizing the code analysis stuff into a single place

PR #70, opened 2024-06-07
Enable ConcurrentExecution and disable generated code analysis

MattKotsenas:
Two simple and related performance improvements:

1. Enable concurrent execution of the analyzer. Verified via manual code inspection + tests that there's no state to maintain between invocations
2. Disable the analyzer in generate code contexts. Things like Razor and Entity Framework can generate large volumes of code that the user isn't directly responsible for.


MattKotsenas:
I don't want to DRY this out via a base class, as it merely _happens to be_ that the generated code configuration is the same for all the analyzers and the behavior is already caught with a test.

Concurrent execution is caught via `RS1026`, which I'd like to enforce via `WarningsAsErrors`.

rjmurillo (approved):

Issue #64, opened 2024-06-06
Investigate why Microsoft.CodeAnalysis.Testing is so slow


MattKotsenas:
The official test harness is slower than our own custom harness. It tests more though, so the comparison isn't apples-to-apples.

This is a tracking bug to investigate the slowness and fix, either on our end or upstream.

MattKotsenas:
Opened a PR upstream to document the reference assemblies caching best practice: https://github.com/dotnet/roslyn-sdk/pull/1160

PR #71, opened 2024-06-07
Cache reference assemblies between tests to improve performance

MattKotsenas:
The testing framework does a bunch of heavy lifting to resolve references for set of `ReferenceAssemblies`, including potentially running the NuGet client to download packages. This class caches the ReferenceAssemblies class (which is thread-safe), so that package resolution only happens once for a given configuration.

The is the same approach taken by https://github.com/dotnet/roslyn-analyzers/blob/4d5fd9da36d64d4c3370b8813122e226844fc6ed/src/Test.Utilities/AdditionalMetadataReferences.cs.

This cuts test time from ~40 seconds to ~20 seconds on my machine.


rjmurillo (approved):

rjmurillo:
Nice!

MattKotsenas:
I also opened a PR to update the test docs to make this clearer: https://github.com/dotnet/roslyn-sdk/pull/1160

MattKotsenas:
This addresses part of #64.

Issue #26, opened 2024-05-31
Enable warnings as errors


MattKotsenas:
- [ ] Turn on warnings as errors
- [ ] Also cleanup `<NoWarn>` elements
- [ ] Clean up warnings backlog

```xml
<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
<MSBuildTreatWarningsAsErrors>true</MSBuildTreatWarningsAsErrors>
```

PR #73, opened 2024-06-07
Update code to remove warnings from backlog

rjmurillo:
This reduces the backlog of warnings by about half, removing many cases for passing null references to methods and null dereference. The remaining warnings are from multiple sources: TODO, diagnostic best practices, and some tricker items we'll need to look at

Related to #26

### DIFF
--- a/Source/Moq.Analyzers/MoqMethodDescriptor.cs
+++ b/Source/Moq.Analyzers/MoqMethodDescriptor.cs
@@ -1,14 +1,15 @@
-﻿using System.Text.RegularExpressions;
+﻿using System.Diagnostics;
+using System.Text.RegularExpressions;
 
 namespace Moq.Analyzers;
 
 internal class MoqMethodDescriptor
 {
-    private readonly bool isGeneric;
+    private readonly bool _isGeneric;
 
     public MoqMethodDescriptor(string shortMethodName, Regex fullMethodNamePattern, bool isGeneric = false)
     {
-        this.isGeneric = isGeneric;
+        _isGeneric = isGeneric;
         ShortMethodName = shortMethodName;
         FullMethodNamePattern = fullMethodNamePattern;
     }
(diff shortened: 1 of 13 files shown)


Source/Moq.Analyzers/MoqMethodDescriptor.cs
```diff
@@ -1,14 +1,15 @@
-﻿using System.Text.RegularExpressions;
+﻿using System.Diagnostics;
+using System.Text.RegularExpressions;
 
 namespace Moq.Analyzers;
 
 internal class MoqMethodDescriptor
 {
-    private readonly bool isGeneric;
+    private readonly bool _isGeneric;
```
rjmurillo, line 8: @MattKotsenas There are some conflicting style rules with this. We can take it up as a separate issue and PR, this is big enough already

Source/Moq.Analyzers/NoConstructorArgumentsForInterfaceMockAnalyzer.cs
```diff
@@ -28,38 +30,44 @@ private static void Analyze(SyntaxNodeAnalysisContext context)
...
         {
-            var qualifiedName = objectCreation.Type as QualifiedNameSyntax;
             genericName = qualifiedName.Right as GenericNameSyntax;
         }
 
         if (genericName?.Identifier == null || genericName.TypeArgumentList == null) return;
 
         // Quick and dirty check
-        if (genericName.Identifier.ToFullString() != "Mock") return;
+        if (!string.Equals(genericName.Identifier.ToFullString(), "Mock", StringComparison.Ordinal)) return;
```
rjmurillo, line 42: If we're going to continue to compare against string literals we should DRY this out

MattKotsenas: Yeah, I think another step to make "soon" is to start to pull things into a set of extension methods on the appropriate syntax or semantic element.

MattKotsenas (approved):

PR #74, opened 2024-06-07
Convert from var to explicit type

### LINKED ISSUE #53: Convert to explicit types
There is mixed use of implicit (`var`) and explicit types within the code base. Since we cannot use implicit types everywhere, and using implicit types makes things difficult during refactoring, switch to explicit types. 

- [ ] Update code to explicit types
- [ ] Set code analysis rule to prevent use of `var` keyword

rjmurillo:
There is mixed use of explicit types and var keyword. There are also cases where the var keyword cannot be used. Rather than have mixed cases, update all code to use explicit types and enable .editorconfig rule to enforce code style

Fixes #53

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -88,9 +88,9 @@ dotnet_remove_unnecessary_suppression_exclusions = none
 [*.cs]
 
 # var preferences
-csharp_style_var_elsewhere = false:error
-csharp_style_var_for_built_in_types = false:error
-csharp_style_var_when_type_is_apparent = false:error
+csharp_style_var_elsewhere = false:warning
+csharp_style_var_for_built_in_types = false:warning
+csharp_style_var_when_type_is_apparent = false:warning
 
 # Expression-bodied members
 csharp_style_expression_bodied_accessors = true:silent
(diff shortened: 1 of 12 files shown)


.editorconfig
```diff
@@ -88,9 +88,9 @@ dotnet_remove_unnecessary_suppression_exclusions = none
 [*.cs]
 
 # var preferences
-csharp_style_var_elsewhere = false:silent
```
MattKotsenas, line 91: So, my experience from making things errors in other projects is that it becomes super annoying during local dev.

Like we're making it impossible to use `var` while hacking on things.

Can we make these warn and then rely on TreatWarningsAsErrors in CI to prevent it from getting checked in?

I know this is a slight reversal from my position a while ago, but I think I'm landing on "style" should be enforced in CI and easily fixable locally, but not block builds.

I think warn is the easiest way to do that.

MattKotsenas (changes requested):

.editorconfig
rjmurillo: I think it's reasonable. We can change it either way if it gives us trouble.

MattKotsenas (approved):

PR #82, opened 2024-06-12
Refactor test cases to matrix on Moq version

### LINKED ISSUE #58: Upgrade (or at least test) with a newer version of Moq
We're currently testing with Moq 4.8.2, which is from 2018. v4.13.1 changed the internal implementation of `.As<T>()` (see https://github.com/devlooped/moq/commit/b552aeddd82090ee0f4743a1ab70a16f3e6d2d11) such that testing with that version or newer makes the `AsAcceptOnlyInterfaces` tests fail. Set a breakpoint on `MoqMethodDescriptor.IsMoqMethod`:

`symbolInfo.Symbol.ToString()`

Before: `"Moq.Mock.As<AsAcceptOnlyInterface.TestBadAsForAbstractClass.BaseSampleClass>()"`
After: `"Moq.Mock<AsAcceptOnlyInterface.TestBadAsForAbstractClass.BaseSampleClass>.As<AsAcceptOnlyInterface.TestBadAsForAbstractClass.BaseSampleClass>()"`

We should:

1. [ ] Fix bug
2. [ ] Consider testing against multiple versions of Moq
3. [ ] Consider enforcing a minimum version?

MattKotsenas:
Refactor test cases so that they test with an "old" version of Moq (4.8.2) and now also a "new" version of Moq (4.18.4).

Fixes #58 

4.8.2 was the pre-existing test version. Any version prior to 4.13.1 works. (4.13.1 changed the internal implementation of `.As<T>()` [see https://github.com/devlooped/moq/commit/b552aeddd82090ee0f4743a1ab70a16f3e6d2d11]).

4.18.4 is the most downloaded version of Moq.

### DIFF
--- a/Source/Moq.Analyzers.Test/AsAcceptOnlyInterfaceAnalyzerTests.cs
+++ b/Source/Moq.Analyzers.Test/AsAcceptOnlyInterfaceAnalyzerTests.cs
@@ -1,3 +1,4 @@
+using Microsoft.CodeAnalysis.Testing;
 using Verifier = Moq.Analyzers.Test.Helpers.AnalyzerVerifier<Moq.Analyzers.AsShouldBeUsedOnlyForInterfaceAnalyzer>;
 
 namespace Moq.Analyzers.Test;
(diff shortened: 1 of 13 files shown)


MattKotsenas:
This will fail until 81 is merged.

rjmurillo (approved):

PR #83, opened 2024-06-12
Update to add devskim

rjmurillo:
Adds [DevSkim](https://github.com/microsoft/DevSkim?tab=readme-ov-file) [rules](https://github.com/microsoft/DevSkim/tree/main/rules/default) to check for correctness and security

- Added as a task on the Linux builder (task only supports Linux) that scans as part of the build. Build is run periodically so we would get new notifications if something gets flagged.
- Added code analysis package so developers are warned as they're writing code of DevSkim issues. This prevents new changes from getting committed to source and shifts left detection to code author time, rather than CI time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -107,8 +107,10 @@ jobs:
 
     - name: Run DevSkim scanner
       uses: microsoft/DevSkim-Action@v1
+      if: runner.os == 'Linux'
 
     - name: Upload DevSkim scan results to GitHub Security tab
       uses: github/codeql-action/upload-sarif@v2
+      if: runner.os == 'Linux'
       with:
         sarif_file: devskim-results.sarif
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -104,3 +104,11 @@ jobs:
       uses: github/codeql-action/analyze@v3
       with:
         category: "/language:csharp"
+
+    - name: Run DevSkim scanner
+      uses: microsoft/DevSkim-Action@v1
+
+    - name: Upload DevSkim scan results to GitHub Security tab
+      uses: github/codeql-action/upload-sarif@v2
+      with:
+        sarif_file: devskim-results.sarif
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -15,6 +15,13 @@
         "dotnet-verify"
       ],
       "rollForward": false
+    },
+    "microsoft.cst.devskim.cli": {
+      "version": "1.0.33",
+      "commands": [
+        "devskim"
+      ],
+      "rollForward": false
     }
   }
 }
--- a/build/targets/codeanalysis/CodeAnalysis.props
+++ b/build/targets/codeanalysis/CodeAnalysis.props
@@ -24,5 +24,9 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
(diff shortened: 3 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -104,3 +104,13 @@ jobs:
       uses: github/codeql-action/analyze@v3
       with:
         category: "/language:csharp"
+
+    - name: Run DevSkim scanner
```
MattKotsenas, line 108: I don't like that there isn't a simple way to repro issues locally. Can we integrate it in some way to make local dev work as well?

## Good

- Add to dotnet tools
- Add `dotnet tool restore` to YAML (sorry, I thought I added it already)
- Add basic pwsh script to run devskim

## Better

- Same as above, but integrate into MSBuild via `Exec` or similar

MattKotsenas: That would also avoid the linux-only issue the GitHub Action has

MattKotsenas (changes requested):

.github/workflows/main.yml
rjmurillo: There's a package we can integrate https://www.nuget.org/packages/Microsoft.CST.DevSkim/

rjmurillo: > That would also avoid the linux-only issue the GitHub Action has

I'm already over the Windows builder. It's. So. Slow.

Zootopia slow sloth stamp

rjmurillo: @MattKotsenas Updated to also include the CA rules, so it's a "yes, and"

MattKotsenas: I think that package is just a lib though: https://nuget.info/packages/Microsoft.CST.DevSkim/1.0.33 so I'm not sure how to drive it from that package. Haven't looked much though so go for it.

rjmurillo: Can't use it anyway, doesn't support netstandard 2.0 (only 2.1). I'll go the global tool route so we have something. There's also a VSIX, but I don't like that for various reasons.

rjmurillo: @MattKotsenas we're going hold off on this. I've raised two issues with DevSkim, https://github.com/microsoft/DevSkim/issues/619 and https://github.com/microsoft/DevSkim/issues/618 to address the issues. 619 is the one I want, so it's super consistent with everything else and "just works"

rjmurillo: Update: The "Good" or "Better" options above are the resolution to https://github.com/microsoft/DevSkim/issues/618

> It seems that a Dockerfile defined action (as DevSkim-Action is) is restricted to Linux pipelines only - the workaround seems to be to migrate to a Typescript action. I found a similar issue reported in another repo on GitHub with the same conclusion - https://github.com/peter-evans/create-pull-request/issues/40. For us, the workaround is a little undesirable here - the docker method allows us to ensure that the .NET SDK is available and pull the latest version of the DevSkim tool, without interfering with the users environment, and ultimately the equivalent behavior of the action is only a couple lines of YML.

Link: https://github.com/microsoft/DevSkim/issues/618#issuecomment-2166879032

I think I'm going to abandon this until https://github.com/microsoft/DevSkim/issues/619 is implemented. Right now all the rules are RegEx so the perf is 1) not good, and 2) not always attached to a symbol. There will need to be some reworking on their end to get this to work as a proper analyzer.

What we're missing out on doesn't seem too critical for this project. Lots of making sure you're not doing silly things like disabling HTTPS checks, making sure you use secure URIs, not using weak crypto algos, vulnerable NuGet packages, and so on. Some of those are covered already by other analyzers (built in or otherwise).

PR #84, opened 2024-06-12
Add analyzer release tracking

### LINKED ISSUE #72: Enable analyzer release tracking
Shows up as RS2008

https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md

Related to #26

MattKotsenas:
Enable analyzer release tracking (i.e. `AnalyzerReleases.Shipped.md`). Fixes #72.

Wokflow was as follows:

1. Download old release from NuGet
2. Open in ILSpy to view the analyzer metadata
3. Create appropriate release in `AnalyzerReleases.Shipped.md`
4. GOTO 1
5. Add help links pointing to our `//docs/rules` files
    - I tried to make the base URL a helper function, but the analyzer is unable to add documentation URLs to the Unshipped file if the URL isn't a constant. So instead I used string interpolation so at least the ID isn't repeated.
6. Use release analyzer codefix to add an unshipped section that adds help links

### DIFF
--- a/Source/Moq.Analyzers/AnalyzerReleases.Shipped.md
+++ b/Source/Moq.Analyzers/AnalyzerReleases.Shipped.md
@@ -50,3 +50,11 @@ Moq1000 | Moq | Warning | NoSealedClassMocksAnalyzer
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
 Moq1002 | Moq | Warning | ConstructorArgumentsShouldMatchAnalyzer
+
+## Release 0.0.7
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+Moq1200 | Moq | Error | SetupShouldBeUsedOnlyForOverridableMembersAnalyzer
--- a/Source/Moq.Analyzers/AnalyzerReleases.Shipped.md
+++ b/Source/Moq.Analyzers/AnalyzerReleases.Shipped.md
@@ -58,3 +58,11 @@ Moq1002 | Moq | Warning | ConstructorArgumentsShouldMatchAnalyzer
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
 Moq1200 | Moq | Error | SetupShouldBeUsedOnlyForOverridableMembersAnalyzer
+
+## Release 0.0.8
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+Moq1300 | Moq | Error | AsShouldBeUsedOnlyForInterfaceAnalyzer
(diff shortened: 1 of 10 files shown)


Source/Moq.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,17 @@
+﻿; Unshipped analyzer release
+; https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
+
+### Changed Rules
+
+Documentation links added for every rule.
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+Moq1000 | Moq | Warning | NoSealedClassMocksAnalyzer, [Documentation](https://github.com/rjmurillo/moq.analyzers/blob/main/docs/rules/Moq1000.md)
```
rjmurillo, line 10: As of 0.0.9 these are all shipped, no?

MattKotsenas: Yeah, changes incoming here.

The idea was that we added the HelpUri, which made it a "changed rule", but the analyzer only cares about (Id, Category, Severity).

Over the releases we've changed which analyzer used which ID, so I'll need to adjust.

rjmurillo: > Over the releases we've changed which analyzer used which ID, so I'll need to adjust.

🤦

rjmurillo (changes requested):

Source/Moq.Analyzers/AnalyzerReleases.Unshipped.md
MattKotsenas: OK, the files are now correct according to the analyzer. However, we have a history issue: Over time we've remapped IDs to analyzers, so adding the documentation links into the release tracking document means either:

1. Adding a link to the _current_ ID, which may not be what the analyzer checked at the time
2. Adding a link to the _original_ ID, which will mean the ID in the document won't match the help link

Thoughts?

Source/Moq.Analyzers/AnalyzerReleases.Shipped.md
```diff
@@ -0,0 +1,76 @@
...
+--------|----------|----------|-------
+Moq1300 | Moq | Error | AsShouldBeUsedOnlyForInterfaceAnalyzer
+
+## Release 0.0.9
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+Moq1201 | Moq | Error | SetupShouldNotIncludeAsyncResultAnalyzer
```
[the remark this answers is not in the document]
MattKotsenas: The analyzer won't allow me to create a changed section for a behavior change. "Changed" to the analyzer specifically means changing the category or the severity (for instance it won't allow me to document that the name of the analyzer itself changed).

rjmurillo (approved):
LGTM. Thanks!

PR #89, opened 2024-06-14
Add more analyzers and bump Meziantou.Analzer to latest version

rjmurillo:
Several changes:

1. Add several new analyzers to check for code correctness and style

- [Sonar](https://www.nuget.org/packages/SonarAnalyzer.CSharp)
- [Multithreading](https://www.nuget.org/packages/Microsoft.VisualStudio.Threading.Analyzers)
- [C# Guidelines](https://www.nuget.org/packages/CSharpGuidelinesAnalyzer)
- [Exhaustive matching](https://www.nuget.org/packages/ExhaustiveMatching.Analyzer)

2. Added suppressions for some of the new rules as they'll need more extensive refactoring to satisfy (e.g., methods have multiple responsibilities, there's too much coupling, poor encapsulation, etc.). Tracked in #90
3. Bump Meziantou.Analyzer to 2.0.157

### DIFF
--- a/build/targets/codeanalysis/CodeAnalysis.props
+++ b/build/targets/codeanalysis/CodeAnalysis.props
@@ -25,5 +25,25 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="SonarAnalyzer.CSharp">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="CSharpGuidelinesAnalyzer">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="ExhaustiveMatching.Analyzer" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(IsTestProject)' == 'true' ">
+    <PackageReference Include="xunit.analyzers">
+      <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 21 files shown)


build/targets/codeanalysis/CodeAnalysis.props
```diff
@@ -25,5 +25,25 @@
...
+    </PackageReference>
+    <PackageReference Include="CSharpGuidelinesAnalyzer">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="ExhaustiveMatching.Analyzer" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(IsTestProject)' == 'true' ">
+    <PackageReference Include="xunit.analyzers">
```
MattKotsenas, line 44: This is pulled in by the main xunit package.

MattKotsenas (changes requested):
We're pretty close to being able to turn on WarningsAsErrors (which I think is valuable) and these add a whole bunch of new items, some of dubious value (do we really need another compiler warning for TODO, or that `f` isn't a good variable name?)

Can we add these and be clean on them at the same time, or suppress violations inline and file TODOs?

build/targets/codeanalysis/CodeAnalysis.props
rjmurillo: Oh, cool. We had this as a separate discrete item before. I'll remove it.

rjmurillo:
> We're pretty close to being able to turn on WarningsAsErrors (which I think is valuable) and these add a whole bunch of new items, some of dubious value (do we really need another compiler warning for TODO, or that `f` isn't a good variable name?)
> 
> Can we add these and be clean on them at the same time, or suppress violations inline and file TODOs?

I can do two things:

1. Wait until you've cleaned up the existing warnings as errors and then rebase
2. For the newly added analyzers, get into a clean state (this is quite reasonable)

build/targets/codeanalysis/CodeAnalysis.props
MattKotsenas: Whoops! Thanks for fixing!

MattKotsenas:
If you get done before me you don't need to wait. I think 2 is more important than 1.

build/targets/codeanalysis/CodeAnalysis.props
rjmurillo: Fixed in 8a63edac

MattKotsenas (approved):

MattKotsenas:
@rjmurillo, this is ready to merge, but the code coverage is blocking merge. Either address, or feel free to lower the threshold for now while we have a lot of churn. I'd be OK with making this informational and not blocking while it's just us.

PR #91, opened 2024-06-14
Enable warnings as errors

rjmurillo:
- Updated `.editorconfig` to set policy on some duplicate warnings for TODO. Borrowed from #89 to make merging easier
- Updated cases where nullability was incorrect (something could be passed as null but wasn't, something was marked as nullable but wasn't). Where possible added asserts so we can catch cases when they arise.
- Removed `<NoWarn>` from `Moq.Analyzers.csproj`
- Added `<TreatWarningsAsErrors>` and `MSBuildTreatWarningsAsErrors` elements to `CodeAnalysis.props`
  - Set each to `true` when MSBuild property `$(ContinuousIntegrationBuild)` is `true` from `DotNet.ReproducibleBuilds` package

### DIFF
--- a/Source/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodCodeFix.cs
+++ b/Source/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodCodeFix.cs
@@ -59,14 +59,10 @@ public class CallbackSignatureShouldMatchMockedMethodCodeFix : CodeFixProvider
     {
         SemanticModel? semanticModel = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
 
-        Debug.Assert(semanticModel != null, nameof(semanticModel) + " != null");
-
-#pragma warning disable S2583 // Conditionally executed code should be reachable
         if (semanticModel == null)
         {
             return document;
         }
-#pragma warning restore S2583 // Conditionally executed code should be reachable
 
         if (oldParameters?.Parent?.Parent?.Parent?.Parent is not InvocationExpressionSyntax callbackInvocation)
         {
--- a/Source/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodCodeFix.cs
+++ b/Source/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodCodeFix.cs
@@ -60,10 +60,12 @@ public class CallbackSignatureShouldMatchMockedMethodCodeFix : CodeFixProvider
 
         Debug.Assert(semanticModel != null, nameof(semanticModel) + " != null");
 
+#pragma warning disable S2583 // Conditionally executed code should be reachable
         if (semanticModel == null)
         {
             return document;
         }
+#pragma warning restore S2583 // Conditionally executed code should be reachable
 
         if (oldParameters?.Parent?.Parent?.Parent?.Parent is not InvocationExpressionSyntax callbackInvocation)
         {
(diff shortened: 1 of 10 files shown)


build/targets/codeanalysis/CodeAnalysis.props
```diff
@@ -5,6 +5,8 @@
     <AnalysisMode>preview</AnalysisMode>
     <WarningLevel>9999</WarningLevel>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
MattKotsenas, line 8: How do you feel about hooking this into `$(IsContinuousIntegration)`, or a new property that defaults to IsContinuousIntegration?

Otherwise, local dev gets hung up on things like incorrect style, unused parameters, etc.

Source/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodCodeFix.cs
```diff
@@ -60,10 +60,12 @@ private async Task<Document> FixCallbackSignatureAsync(SyntaxNode root, Document
 
         Debug.Assert(semanticModel != null, nameof(semanticModel) + " != null");
 
+#pragma warning disable S2583 // Conditionally executed code should be reachable
```
MattKotsenas, line 63: Why suppress these? I'm not sure of the actual scenario, but based on the nullability annotations, it seems that a null `SemanticModel` is possible. Instead, should these be an early return without the `Debug.Assert`?

build/targets/codeanalysis/CodeAnalysis.props
rjmurillo: I'm fine with that. You want to open a new PR for that, or put it in this one?

Source/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodCodeFix.cs
rjmurillo: I think this is actually a bug with the analyzer. It's theoretically reachable since the `document.GetSemanticModelAsync` can return a `SemanticModel?`, though I don't understand under what circumstance we could not be reachable. I have the assert in there to trip when that would be the case, but I can't actually get it to trip without doing some weird things in the harness.

build/targets/codeanalysis/CodeAnalysis.props
rjmurillo: I pushed df2af35 with the flag

Source/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodCodeFix.cs
rjmurillo: Only times where `SemanticModel` would be null are if the document does not support syntax tree or the project does not support compilation.

MattKotsenas: I'm on my phone, but I have a feeling the Debug.Assert is what's tripping up the analyzer. If you remove it, does the analyzer behave correctly?

rjmurillo: It is; however, it's broken because it trips even in release mode where that won't fire. The examples in their documentation are also not this case https://rules.sonarsource.com/csharp/RSPEC-2583/.

In the case of DEBUG, I _want_ it to fire so I know when we trip over it.

I can put a comment in for _future maintainer_ to know why this is the way it is

MattKotsenas: But why do you want to know when it fires? What do you plan on doing about it? I think it's a valid scenario where the only thing we can reasonably do is early return.

I think the debug asserts should just be removed entirely. Let me know if I'm misunderstanding the scenario / use case.

rjmurillo: Removed in db75e30c

rjmurillo:
@MattKotsenas we can turn off Codacy when we get reporting of coverage up as something we can gate on. We can also check in [a config file](https://docs.codacy.com/repositories-configure/codacy-configuration-file/) to keep the integration sane. I disabled the Sonar analyzer, so I'm really just interested in code coverage and some obvious linting things in YAML, PowerShell, and JSON files. The recommendation from Codacy is to check in the configuration for the tools used themselves and it'll pick them up.

MattKotsenas (approved):

PR #107, opened 2024-06-21
Move code under src/ and tests/ directories

MattKotsenas:
I tried to avoid making this change, but the upcoming Benchmarks PR needs to share an .editorconfig with the unit tests project, so reorganizing the source so that:

1. We follow the common pattern of `src/` and `tests/`
2. Test projects share a common ancestor


rjmurillo (approved):
Thanks for using variables instead of relative dot paths

PR #108, opened 2024-06-21
Update labeler.yml path for source

rjmurillo:
#107 moved the source folder to follow convention. Updating the labeler so PRs are correctly labeled


rjmurillo:
I'm like 50% sure this will work based on the examples https://github.com/actions/labeler/tree/v4. Their test fixtures have indented spaces. It seems to be complaining about the `#` but that's also in their documentation with a preceding line break.

MattKotsenas (approved):

PR #109, opened 2024-06-21
Add Moq.Analyzers.Benchmarks with a single sample benchmark

MattKotsenas:
Add the project `Moq.Analyzers.Benchmarks` with a sample benchmark for diagnostic Moq1300. This pattern is copied from https://github.com/dotnet/roslyn-analyzers/blob/f1115edce8633ebe03a86191bc05c6969ed9a821/src/PerformanceTests/Tests/Enabled/CSharp_CA1416.cs, but where possible I used the types exposed by the Microsoft.CodeAnalysis.*.Testing libraries instead.

As part of the exercise, I filed https://github.com/dotnet/roslyn-sdk/issues/1165 to discuss if some of these helpers should move into the testing packages, and / or to document any existing profiling best practices.

This change doesn't attempt to hook the benchmarks into CI in any way; they're both currently too slow for that, and I'm not yet sure what (if anything) to gate on.


rjmurillo (approved):
Step 2 would be getting this to show up in the action and building gates. Thanks for also taking care of some of those other warnings

MattKotsenas:
Here's a sample of the output:

```
| Method                 | Mean     | Error    | StdDev   | Median   | Ratio | RatioSD | Gen0      | Gen1      | Allocated | Alloc Ratio |
|----------------------- |---------:|---------:|---------:|---------:|------:|--------:|----------:|----------:|----------:|------------:|
| Moq1300WithDiagnostics | 378.6 ms | 12.70 ms | 37.26 ms | 371.9 ms |  1.80 |    0.29 | 9000.0000 | 3000.0000 |  58.72 MB |        1.37 |
| Moq1300Baseline        | 215.3 ms | 11.45 ms | 33.75 ms | 205.8 ms |  1.00 |    0.00 | 7000.0000 | 2000.0000 |  42.95 MB |        1.00 |
```

PR #115, opened 2024-06-24
Add messy implementation to check for default ctor

rjmurillo:


tests/Moq.Analyzers.Test/ConstructorArgumentsShouldMatchAnalyzerTests.cs
```diff
@@ -35,9 +35,15 @@ public static IEnumerable<object[]> TestData()
             ["""new Mock<AbstractGenericClassDefaultCtor<object>>{|Moq1002:(42)|};"""],
             ["""new Mock<AbstractGenericClassDefaultCtor<object>>();"""],
             ["""new Mock<AbstractGenericClassDefaultCtor<object>>(MockBehavior.Default);"""],
-
-            // TODO: "I think this _should_ fail, but currently passes. Tracked by #55."
-            // ["""new Mock<AbstractClassWithCtor>();"""],
+            ["""new Mock<AbstractClassWithCtor>{|Moq1002:()|};"""],
+            ["""new Mock<AbstractClassWithCtor>{|Moq1002:(MockBehavior.Strict)|};"""],
+            ["""new Mock<AbstractClassWithCtor>{|Moq1002:(MockBehavior.Loose)|};"""],
```
MattKotsenas, line 40: nit: Does `Strict` and `Loose` actually test different things? Can we pick one to reduce test cases?

MattKotsenas (approved):

tests/Moq.Analyzers.Test/ConstructorArgumentsShouldMatchAnalyzerTests.cs
rjmurillo: From the analyzer's perspective, no. It cares mostly about the varargs used. Depending on the ctor used, it's either ordinal 0 or 1. There's no difference which enum value is used. Fix tracked in #123

PR #120, opened 2024-06-25
Update main.yml to enable NuGet cache and cached dotnet setup

rjmurillo:


build/targets/compiler/Compiler.props
```diff
@@ -3,6 +3,9 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <LangVersion>12.0</LangVersion>
     <Nullable>enable</Nullable>
+
+    <!-- Avoid NU1403 error when running dotnet restore -->
+    <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
```
MattKotsenas, line 8: This should be set via the reproducible-builds package: https://github.com/dotnet/reproducible-builds/blob/7ae5c7db58eff2a57b556e3f56b2d5e5143518d5/src/DotNet.ReproducibleBuilds.Isolated/Sdk/Sdk.props#L50

Can you see if that package isn't being included correctly?

PR #125, opened 2024-06-25
Refactor Moq1300 to use IOperation-based analysis

MattKotsenas:
Refactor Moq1300 to use `IOperation`-based analysis instead of syntax analysis. This method results in a slightly less specific diagnostic (it uses the location of the entire `.As<T>()` method instead of only the type parameter, however it has several benefits:

1. About double the performance / speed
2. Cuts the memory allocations in half
3. I believe a bit more robust to "interesting" syntax trees

Here's a BenchmarkDotNet run comparing the old and new analyzers:
```
| Method                    | Mean     | Error   | StdDev   | Ratio | RatioSD | Gen0      | Gen1      | Gen2      | Allocated | Alloc Ratio |
|-------------------------- |---------:|--------:|---------:|------:|--------:|----------:|----------:|----------:|----------:|------------:|
| OldMoq1300WithDiagnostics | 204.2 ms | 5.01 ms | 14.44 ms |  1.72 |    0.13 | 8000.0000 | 4000.0000 | 1000.0000 |  69.35 MB |        1.36 |
| NewMoq1300WithDiagnostics | 162.3 ms | 3.21 ms |  3.15 ms |  1.38 |    0.06 | 6000.0000 | 2000.0000 |         - |  62.42 MB |        1.22 |
| Moq1300Baseline           | 118.0 ms | 2.35 ms |  5.12 ms |  1.00 |    0.00 | 5000.0000 | 2000.0000 |         - |  50.97 MB |        1.00 |
```


rjmurillo (dismissed):

MattKotsenas:
I should switch the first or default with a single or default for safety

src/Moq.Analyzers/AsShouldBeUsedOnlyForInterfaceAnalyzer.cs
```diff
@@ -29,43 +29,58 @@ public override void Initialize(AnalysisContext context)
...
-        TypeSyntax typeArgument = typeArguments[0];
-        SymbolInfo symbolInfo = context.SemanticModel.GetSymbolInfo(typeArgument, context.CancellationToken);
-
-        if (symbolInfo.Symbol is ITypeSymbol { TypeKind: not TypeKind.Interface })
+        if (typeArguments[0] is ITypeSymbol { TypeKind: not TypeKind.Interface })
         {
-            context.ReportDiagnostic(Diagnostic.Create(Rule, typeArgument.GetLocation()));
+            // Try to locate the type argument in the syntax tree to report the diagnostic at the correct location.
+            // If that fails for any reason, report the diagnostic on the operation itself.
+            NameSyntax? memberName = context.Operation.Syntax.DescendantNodes().OfType<MemberAccessExpressionSyntax>().Select(mae => mae.Name).FirstOrDefault();
```
rjmurillo, line 80: Note from @MattKotsenas : convert to `SingleOrDefault()`

MattKotsenas: fixed!

src/Moq.Analyzers/EnumerableExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    public static TSource? SingleWhenOnly<TSource>(this IEnumerable<TSource> source)
+    {
+        return source.SingleWhenOnly(_ => true);
+    }
+
+    /// <summary>
+    /// Returns the only element of a sequence that satisfies a specified condition,
+    /// and returns a default value if no such element exists or there are multiple matches.
+    /// </summary>
+    /// <typeparam name="TSource">The type of the <paramref name="source"/> collection.`</typeparam>
```
rjmurillo, line 15: There's an extra ` at the end that's tripping SA1629

src/Moq.Analyzers/EnumerableExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <param name="source">The collection to enumerate.</param>
+    /// <param name="predicate">A function to test each element for a condition.</param>
+    /// <returns><see langword="true"/> if there is only one element in the collection; <see langword="false"/> otherwise.</returns>
```
rjmurillo, line 18: The returns doc is wrong, you return an instance of `TSource` or `default`

src/Moq.Analyzers/EnumerableExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <remarks>
+    /// This should be equivalent to calling <see cref="Enumerable.SingleOrDefault{TSource}(IEnumerable{TSource}, Func{TSource, bool})"/>
+    /// combined with a catch that returns <see langword="null"/>.
+    /// </remarks>
+    public static TSource? SingleWhenOnly<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate)
```
rjmurillo, line 23: I said I was good with the name, but I changed my mind. Sorry about that. Now that I see the code, this feels like `SingleOrDefaultWhenOnly`, except it doesn't really (because we expect a `.Single()` to throw when there are multiple.  

A `.FirstOrDefault()` would look something like

```csharp
foreach(TSource element in source)
{
  if (predicate(element) return element;
}
```

But this is a first or default when there are multiple, so more like Single than First. This extension method needs to have really good documentation around it

src/Moq.Analyzers/EnumerableExtensions.cs
rjmurillo, line 18:
```suggestion
/// <summary>
/// Returns the only element of a sequence that satisfies a specified condition or default if no such element exists or more than one element satisfies the condition.
/// </summary>
/// <param name="source">The collection to enumerate.</param>
/// <param name="predicate">A function to test each element for a condition.</param>
/// <returns>The single element that satisfies the condition, or default if no such element exists or more than one element satisfies the condition.</returns>
```

rjmurillo (changes requested):

src/Moq.Analyzers/EnumerableExtensions.cs
MattKotsenas: from chat:

Decided to go with `DefaultIfNotSingle()`. It doesn't have quite the same discoverability in IntelliSense because it doesn't start with "single", but the name is a lot clearer.

src/Moq.Analyzers/EnumerableExtensions.cs
MattKotsenas: Whoops! Fixed!

src/Moq.Analyzers/EnumerableExtensions.cs
MattKotsenas: Whoops! Fixed

src/Moq.Analyzers/EnumerableExtensions.cs
MattKotsenas: Fixed!

src/Moq.Analyzers/EnumerableExtensions.cs
MattKotsenas: Fixed!

rjmurillo (approved):

PR #137, opened 2024-06-27
Update to latest version of DotNet.ReproducibleBuilds

MattKotsenas:
Update to the latest version of `DotNet.ReproducibleBuilds`. Doing so exposes a difference in how SourceLink is invoked between local and CI scenarios. I make local build work like CI by adding a RepositoryBranch if one isn't supplied (bug filed against the SourceLink repo to do the fallback directly). This then requires pulling in a newer version of `Verify.Nupkg` in order to scrub the new `branch` attribute.

### DIFF
--- a/build/targets/reproducible/Packages.props
+++ b/build/targets/reproducible/Packages.props
@@ -1,5 +1,5 @@
 <Project>
   <ItemGroup>
-    <PackageVersion Include="DotNet.ReproducibleBuilds" Version="1.1.1" />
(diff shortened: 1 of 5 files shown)


MattKotsenas:
The new version of SourceLink that the package brings in adds a `pr` attribute to the source control element of the manifest. This will require an update to the version of Verify.Nupkg first.

rjmurillo (approved):

PR #174, opened 2024-08-07
Update SquiggleCop and remove ErrorLog property

MattKotsenas:
- Update SquiggleCop to the latest version
- Remove the `SetErrorLog` target now that SquiggleCop does so automatically


rjmurillo (approved):
Thanks @MattKotsenas

Issue #172, opened 2024-08-05
Moq1100 incorrectly firing on nullable parameters


marcovr:
Analyzer Moq1100 is incorrectly firing in the following case:
```cs
public interface IFoo
{
    bool DoSomething(object? bar);
}

var mock = new Mock<IFoo>();
mock.Setup(m => m.DoSomething(It.IsAny<object?>())).Returns((object? bar) => true);
```

The analyzer reports:
> Moq1100: Callback signature must match the signature of the mocked method

Note: this not only happens with `object` but also other types

rjmurillo:
@marcovr Thank you for reporting this issue. Would you be able to try a private to ensure this issue is corrected? [Moq.Analyzers.0.2.0-g7dd22517f7.nupkg](https://github.com/rjmurillo/moq.analyzers/actions/runs/10566077819/artifacts/1856241340)

marcovr:
Thanks @rjmurillo!
I can confirm that the preview package resolves the issue 👍🏻.

However, I noticed that it introduced another (possibly unintended?) change in behaviour. Consider the following example:
```cs
var mock = new Mock<IFoo>();
mock.Setup(m => m.DoSomething(42)).Returns((long bar) => true);

public interface IFoo
{
    bool DoSomething(long bar);
}
```

In this case, the analyzer now reports `Moq1100` unless I either perform a cast:
```cs
mock.Setup(m => m.DoSomething((long)42)).Returns((long bar) => true);
```
or change the signature of the return lambda:
```cs
mock.Setup(m => m.DoSomething(42)).Returns((int bar) => true);
```

I don't really like the variant with a cast as it causes other analyzers to report an unnecessary cast. And the other variant works fine, but seems kind of backwards to me. Because I now need a "wrong" signature to make the analyzer happy, which checks for correct signatures 😁

rjmurillo:
Thanks @marcovr for your patience and the additional test cases; I've added them to the suite.

You're right, the updated type conversion detection method caused a regression. In the case there was an `Int32` value being passed to `Int64`, which was failing the now stricter type check. I've updated the implementation since there is a conversion that can happen in a few scenarios:

```csharp
var mock = new Mock<IFoo>();
mock.Setup(m => m.DoSomething(42)).Returns((long bar) => true);
mock.Setup(m => m.DoSomethingElse(42)).Returns((long var) => true);

public interface IFoo
{
    bool DoSomething(long bar);
    bool DoSomethingElse(object bar);
[...]

PR #187, opened 2024-08-26
Moq1100 incorrectly firing on nullable parameters

rjmurillo:
Update type comparison to use `ClassifyConversion` to check if the types are assignable, rather than stringly strictly equal. 

## Changes:

- Add check for types based on assignability (which also covers user-defined implicit and explicit conversions), rather than strictly equal with types. 
- Add logging for test type to aid in debugging cases
- Update detection (span for diagnostic is tighter now)
- Add repro for issue reported in #172 (thanks @marcovr!)
- Add test cases for user-defined implicit and explicit conversions (thanks @MattKotsenas!)

### DIFF
--- a/src/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodAnalyzer.cs
+++ b/src/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodAnalyzer.cs
@@ -128,6 +128,7 @@ public class CallbackSignatureShouldMatchMockedMethodAnalyzer : DiagnosticAnalyz
         {
             case true:
                 return true;
+
             // Fall back to seeking an implicit conversion operator
             // The Conversion type will not show a user defined implicit conversion, so we have to seek it out explicitly
             case false:
--- a/src/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodAnalyzer.cs
+++ b/src/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodAnalyzer.cs
@@ -83,8 +83,12 @@ public class CallbackSignatureShouldMatchMockedMethodAnalyzer : DiagnosticAnalyz
                 continue;
             }
 
-            TypeInfo lambdaParameterType = context.SemanticModel.GetTypeInfo(lambdaParameterTypeSyntax, context.CancellationToken);
-            TypeInfo mockedMethodArgumentType = context.SemanticModel.GetTypeInfo(mockedMethodArguments[argumentIndex].Expression, context.CancellationToken);
+            TypeInfo lambdaParameterType = context.SemanticModel.GetTypeInfo(
+                lambdaParameterTypeSyntax,
+                context.CancellationToken);
+            TypeInfo mockedMethodArgumentType = context.SemanticModel.GetTypeInfo(
+                mockedMethodArguments[argumentIndex].Expression,
+                context.CancellationToken);
 
             ITypeSymbol? lambdaParameterTypeSymbol = lambdaParameterType.Type;
             ITypeSymbol? mockedMethodTypeSymbol = mockedMethodArgumentType.Type;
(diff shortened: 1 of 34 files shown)


src/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodAnalyzer.cs
```diff
@@ -58,31 +58,52 @@ private static void Analyze(SyntaxNodeAnalysisContext context)
...
+            // Check if there's an implicit conversion or identity conversion
+            Conversion conversion = context.SemanticModel.Compilation.ClassifyConversion(lambdaParameterTypeSymbol, mockedMethodTypeSymbol);
 
-                if (!string.Equals(mockedMethodTypeName, lambdaParameterTypeName, StringComparison.Ordinal))
-                {
-                    Diagnostic diagnostic = callbackLambda.ParameterList.CreateDiagnostic(Rule);
-                    context.ReportDiagnostic(diagnostic);
-                }
+            // Where a conversion exists, even if it is explicit
+            if (conversion.Exists && conversion is not { IsImplicit: false, IsIdentity: false, IsExplicit: false })
```
MattKotsenas, line 100: I don't understand this part (and may be my own misunderstanding).

If there's an implicit or identity conversion, then continue, but if there's an explicit conversion available, then don't you also need to check if the cast is being performed here?

MattKotsenas: Also, this code doesn't match the comment on line 96.

tests/Moq.Analyzers.Test/CallbackSignatureShouldMatchMockedMethodCodeFixTests.cs
```diff
@@ -91,6 +119,10 @@ internal interface IFoo
                 int Do(int i, string s, DateTime dt);
 
                 int Do(List<string> l);
+
+                bool Do(object? bar);
+
+                // bool Do(long bar);
```
MattKotsenas, line 125: nit: comment

rjmurillo: Fixed

src/Moq.Analyzers/CallbackSignatureShouldMatchMockedMethodAnalyzer.cs
```diff
@@ -58,31 +58,101 @@ private static void Analyze(SyntaxNodeAnalysisContext context)
...
+                return true;
 
-                if (!string.Equals(mockedMethodTypeName, lambdaParameterTypeName, StringComparison.Ordinal))
+            // Fall back to seeking an implicit conversion operator
+            // The Conversion type will not show a user defined implicit conversion, so we have to seek it out explicitly
+            case false:
                 {
-                    Diagnostic diagnostic = callbackLambda.ParameterList.CreateDiagnostic(Rule);
-                    context.ReportDiagnostic(diagnostic);
+                    foreach (ISymbol? member in destination.GetMembers())
```
MattKotsenas, line 136: This doesn't feel right to me. `Conversion` has an `IsUserDefined` method on it, why doesn't that work?

rjmurillo: Not sure why that doesn't work. Given the following scenario

```csharp
new Mock<IFoo>().Setup(x => x.Do((CustomType)"42")).Returns((CustomType i) => true);

internal interface IFoo
{
    bool Do(CustomType custom);
}

public class CustomType
{
    public int Value { get; }

    public CustomType(int value)
    {
        Value = value;
    }

    // User-defined conversions
    public static implicit operator CustomType(int value)
    {
        return new CustomType(value);
    }

    public static explicit operator CustomType(string str)
    {
        return new CustomType(int.Parse(str));
    }
}
```

The property for `conversion.IsUserDefined` is false. The only things that are true is `Exists`, `IsIdentity` and `IsImplicit`.

In the case `new Mock<IFoo>().Setup(x => x.Do(42)).Returns((CustomType i) => true);` `conversion` has `Explicit`, `IsUserDefined` and `IsImplicit` true.

MattKotsenas (approved):

PR #215, opened 2024-10-06
Use dotnet template for .gitattributes

MattKotsenas:
Use the upcoming (in .NET 10) `.gitattributes` template. The goal is to get some usage of the template before shipping it more broadly.

This PR has two user-visible changes:

1. It normalizes some previously unnormalized files to have `\n` in the git index, as is best practice when working in mixed OSes
2. It removes the forced use of `\n` in the working directory, so that Windows users can use `\r\n` and have them normalized on check-in

However, git doesn't eagerly update working tree line endings. Instead, it updates them on checkout, which requires updated content. If you want to force git to update all working tree line endings, you can run a command such as `git rm --cached -r . && git reset --hard` to re-checkout everything.

## Summary by CodeRabbit

- **Chores**
	- Updated the `.gitattributes` file for improved clarity and handling of various file types.
	- Revised the `.gitignore` file to reflect modern development practices and tools.
	- Reformatted the `LICENSE` file for consistent indentation without altering the legal text.
	- Replaced content in the `Resources.resx` file while maintaining its structure and schema.


rjmurillo (changes requested):
The updated gitignore file is ignoring things like .cmd files. I'll need to look at this more closely

rjmurillo (approved):

PR #216, opened 2024-10-08
Refactor to support RS1038

rjmurillo:
Update code structure to be compliant with [RS1038](https://github.com/dotnet/roslyn-analyzers/blob/main/docs/rules/RS1038.md) related to #214 

## Changes

- Created a new project `Moq.CodeFixes` to encapsulate only the code fixes
- Removed the `Microsoft.CodeAnalysis.CSharp.Workspaces` reference on `Moq.Analyzers` causing analyzer RS1038 to trigger, moved to `Moq.CodeFixes`
- Linked shared files between projects
- Minor formatting changes to align with checked in standards

## Summary by CodeRabbit

- **New Features**
	- Introduced a new project for `Moq.CodeFixes`, enhancing the functionality of the Moq analyzers.
	- Added support for new code fixes related to callback signature matching.

- **Bug Fixes**
	- Updated callback signature tests to ensure they accurately reflect expected method signatures.

- **Documentation**
	- Improved clarity in code comments and class summaries for better understanding.

- **Chores**
	- Updated package references for better dependency management.
	- Added project references to facilitate integration testing of code fixes.


Directory.Packages.props
```diff
@@ -16,8 +16,7 @@
       - https://github.com/dotnet/roslyn/blob/main/docs/wiki/NuGet-packages.md
       - https://learn.microsoft.com/en-us/visualstudio/productinfo/vs-servicing
     -->
-
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.3.1" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.8.0" />
```
MattKotsenas, line 19: This means that our users must be running .NET SDK 8+. Do we want to force that before .NET 6 is out-of-support?

tests/Moq.Analyzers.Test/Moq.Analyzers.Test.csproj
```diff
@@ -26,6 +26,7 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)/src/Moq.Analyzers/Moq.Analyzers.csproj" AddPackageAsOutput="true" />
+    <ProjectReference Include="..\..\src\Moq.CodeFixes\Moq.CodeFixes.csproj" />
```
MattKotsenas, line 29: I don't think you should need to do this? Shouldn't it be brought in via the reference in analyzers?

tests/Moq.Analyzers.Benchmarks/Moq.Analyzers.Benchmarks.csproj
```diff
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)/src/Moq.Analyzers/Moq.Analyzers.csproj" />
+    <ProjectReference Include="..\..\src\Moq.CodeFixes\Moq.CodeFixes.csproj" />
```
MattKotsenas, line 11: I don't think you should need to do this? Shouldn't it be brought in via the reference in analyzers?

MattKotsenas (changes requested):
I would expect to see a new .dll referenced in `tests/Moq.Analyzers.Test/PackageTests.Baseline#contents.verified.txt`, but don't. That makes me think that the code fixes aren't in the package.

Directory.Packages.props
rjmurillo: This is now back to 4.3.1 and both `CodeAnalysis.CSharp` and the `Workspace` packages are next to each other under the comment so future me doesn't make the same mistake

tests/Moq.Analyzers.Benchmarks/Moq.Analyzers.Benchmarks.csproj
rjmurillo: Yes, it's transitively included. I've removed it in the update for the benchmark and test projects

MattKotsenas (approved):

PR #226, opened 2024-10-15
Add rule to explicitly pick MockBehavior

### LINKED ISSUE #78: Add analyzer for default `MockBehavior`
Many devs have strong opinions on `MockBehavior.Strict`, `MockBehavior.Loose` and the default behavior (Loose).

We should find a way to balance letting devs enforce the rules they want while not enforcing a particular opinion.

Designs to consider

1. An analyzer that flags when no behavior is specified or the default enum is used. This wouldn't enforce Loose vs Strict, just require people to use either Loose or Strict for clarity. I think this should be enabled by default as a warning.

2. A rule that requires Strict mocking. This would flag Loose, Default, or no specification. This should either be off by default, or on at Info level and devs can change the default using .editorconfig or a .global config file.

Related to #127

MattKotsenas:
Fixes #78 

Add rule "Moq1400: Explicitly choose a mocking behavior instead of relying on the default (Loose) behavior"

This addresses a top pain point for Moq in large teams: accidentally relying on Moq's default "loose" behavior. Not specifying a behavior or specifying `MockBehavior.Default` in `new Mock()`, `Mock.Of()`, or `new MockRepository()` raises a warning.

This change does _not_ enforce `MockBehavior.Strict` everywhere. That will be a separate rule (coming later) as it's more controversial and thus likely requires some configuration.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new rule, `Moq1400`, to enforce explicit selection of mocking behavior in the Moq framework.
	- Added a diagnostic analyzer to alert developers when default mock behavior is used without explicit specification.

- **Bug Fixes**
	- Enhanced validation to ensure that the `Mock.As()` method is only used with interface types.

- **Tests**
	- Added unit tests for the `SetExplicitMockBehaviorAnalyzer` to verify its functionality in detecting missing explicit mock behavior.

- **Documentation**
	- Updated documentation for the new rule and provided examples for correct usage.

### DIFF
--- a/src/Moq.Analyzers/SetExplicitMockBehaviorAnalyzer.cs
+++ b/src/Moq.Analyzers/SetExplicitMockBehaviorAnalyzer.cs
@@ -91,7 +91,7 @@ public class SetExplicitMockBehaviorAnalyzer : DiagnosticAnalyzer
             if (argument.Value is IFieldReferenceOperation fieldReferenceOperation)
             {
                 ISymbol field = fieldReferenceOperation.Member;
-                if (field.ContainingType.IsInstanceOf(mockBehaviorSymbol) && IsNotDefaultBehavior(field.Name))
+                if (field.ContainingType.IsInstanceOf(mockBehaviorSymbol) && IsExplicitBehavior(field.Name))
                 {
                     return;
                 }
@@ -119,7 +119,7 @@ public class SetExplicitMockBehaviorAnalyzer : DiagnosticAnalyzer
             if (argument.Value is IFieldReferenceOperation fieldReferenceOperation)
             {
                 ISymbol field = fieldReferenceOperation.Member;
-                if (field.ContainingType.IsInstanceOf(mockBehaviorSymbol) && IsNotDefaultBehavior(field.Name))
+                if (field.ContainingType.IsInstanceOf(mockBehaviorSymbol) && IsExplicitBehavior(field.Name))
                 {
                     return;
                 }
@@ -129,7 +129,7 @@ public class SetExplicitMockBehaviorAnalyzer : DiagnosticAnalyzer
         context.ReportDiagnostic(invocationOperation.Syntax.GetLocation().CreateDiagnostic(Rule));
     }
 
-    private static bool IsNotDefaultBehavior(string symbolName)
+    private static bool IsExplicitBehavior(string symbolName)
     {
         return string.Equals(symbolName, "Loose", StringComparison.Ordinal) || string.Equals(symbolName, "Strict", StringComparison.Ordinal);
     }


src/Moq.Analyzers/SetExplicitMockBehaviorAnalyzer.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            return;
+        }
+
+        // Look for the MockBehavior type and provide it to Analyze to avoid looking it up multiple times.
+        INamedTypeSymbol? mockBehaviorSymbol = context.Compilation.GetTypeByMetadataName(WellKnownTypeNames.MoqBehavior);
+        if (mockBehaviorSymbol is null)
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
MattKotsenas: Incorrect. Part of global usings.

rjmurillo (dismissed):
Nit: Explicit is missing the first 'i'; otherwise, LGTM 👍

src/Moq.Analyzers/SetExplicitMockBehaviorAnalyzer.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        foreach (IArgumentOperation argument in creationOperation.Arguments)
+        {
+            if (argument.Value is IFieldReferenceOperation fieldReferenceOperation)
+            {
+                ISymbol field = fieldReferenceOperation.Member;
+                if (field.ContainingType.IsInstanceOf(mockBehaviorSymbol) && IsNotDefaultBehavior(field.Name))
```
MattKotsenas, line 94:
```suggestion
                if (field.ContainingType.IsInstanceOf(mockBehaviorSymbol) && IsExplicitBehavior(field.Name))
```

src/Moq.Analyzers/SetExplicitMockBehaviorAnalyzer.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            return;
+        }
+
+        foreach (IArgumentOperation argument in invocationOperation.Arguments)
+        {
+            if (argument.Value is IFieldReferenceOperation fieldReferenceOperation)
+            {
+                ISymbol field = fieldReferenceOperation.Member;
+                if (field.ContainingType.IsInstanceOf(mockBehaviorSymbol) && IsNotDefaultBehavior(field.Name))
```
MattKotsenas, line 122:
```suggestion
                if (field.ContainingType.IsInstanceOf(mockBehaviorSymbol) && IsExplicitBehavior(field.Name))
```

PR #236, opened 2024-10-23
Remove ECS0200 suppression and update WellKnownTypes to not trigger ECS0600

rjmurillo:
Effective C# rules 0200 and 0600 have violations in `src/Common/WellKnownTypeNames.cs`. This change updates the members in `WellKnownTypes` type and their usages

Related to https://github.com/rjmurillo/EffectiveCSharp.Analyzers/issues/73, see item for discussion

- Updated constant declarations to static read-only fields for improved flexibility as referenced in [ECS0200](https://github.com/rjmurillo/EffectiveCSharp.Analyzers/blob/main/docs/rules/ECS0200.md) and removed suppression.
- Updated naming conventions for `WellKnownTypes` to avoid triggering [ECS0600](https://github.com/rjmurillo/EffectiveCSharp.Analyzers/blob/main/docs/rules/ECS0600.md)
- Updated control flow in `src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs` to move from switch, which requires compile time constants, to a more traditional if/else
- Updated names in the former `WellKnownTypeNames` type (now `WellKnownMoqNames`) with improved names and added documentation


src/Common/WellKnownTypeNames.cs
```diff
@@ -1,18 +1,16 @@
...
-    internal const string MockBehavior = nameof(MockBehavior);
-    internal const string MockFactory = nameof(MockFactory);
-    internal const string MoqMock = $"{Moq}.{MockName}";
-    internal const string MoqMock1 = $"{MoqMock}`1";
-    internal const string MoqBehavior = $"{Moq}.{MockBehavior}";
-    internal const string MoqRepository = $"{Moq}.MockRepository";
-    internal const string As = nameof(As);
-    internal const string Create = nameof(Create);
-    internal const string Of = nameof(Of);
+    internal static readonly string Moq = nameof(Moq);
```
MattKotsenas, line 5: The use of `nameof()` in these cases is wrong. These are aren't "true" constants like pi, but they are in the sense that renaming this field shouldn't change the value.

MattKotsenas (changes requested):

src/Common/WellKnownTypeNames.cs
rjmurillo: The use in the code base is dual purpose: 

one is the namespace (which is probably as close as you're going to get to being constant), 

https://github.com/rjmurillo/moq.analyzers/blob/bbb213df094b4bf52314a5a066a9c775280ea013/src/Common/MoqMethodDescriptorBase.cs#L14

the other is for a type name

https://github.com/rjmurillo/moq.analyzers/blob/bbb213df094b4bf52314a5a066a9c775280ea013/src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs#L300

The convention in WellKnownTypeName.cs can be updated to make the use clearer and avoid tripping either rule, example:

```suggestion
    internal static readonly string MoqNamespace = "Moq";
    internal static readonly string MoqSymbolName = "Moq";
```

MattKotsenas: I'm happy with a change like that

rjmurillo: In a4e7819b8b24adeb5e0eec7537b974f04c85b89f

src/Common/WellKnownTypeNames.cs
```diff
@@ -1,18 +1,17 @@
...
-    internal const string MoqMock = $"{Moq}.{MockName}";
-    internal const string MoqMock1 = $"{MoqMock}`1";
-    internal const string MoqBehavior = $"{Moq}.{MockBehavior}";
-    internal const string MoqRepository = $"{Moq}.MockRepository";
-    internal const string As = nameof(As);
-    internal const string Create = nameof(Create);
-    internal const string Of = nameof(Of);
+    internal static readonly string MoqNamespace = "Moq";
+    internal static readonly string MoqSymbolName = "Moq";
+    internal static readonly string Mock = nameof(Mock);
```
MattKotsenas, line 7: What's the rationale for `Mock`, `As`, `Create` and `Of` not also being converted? To me they're all in the same category.

MattKotsenas (changes requested):

MattKotsenas (approved):

PR #244, opened 2024-10-25
Disable SquiggleCop for PerfDiff

MattKotsenas:
To streamline the development workflow I've disabled SquiggleCop for the PerfDiff project.

PerfDiff isn't "shipping" code, and it's a frequent source of missed SquiggleCop updates because it isn't part of the `sln`.


rjmurillo (approved):
https://github.com/rjmurillo/moq.analyzers/blob/main/build/scripts/Set-SquiggleCopBaseline.ps1 was created to update all csproj files regardless if they're in the sln

PR #245, opened 2024-10-27
Use WellKnownTypeProvider + KnownSymbols pattern to simplify analyzers

MattKotsenas:
Change adds two related types and refactors the analyzers such that the string-based `WellKnownMockTypes` can be deleted. These changes unlock additional refactoring, but I stopped to get feedback.

The main changes are:

## WellKnownTypeProvider

WellKnownTypeProvider comes from [roslyn-analyzers](https://github.com/[dotnet/roslyn-analyzers](https://github.com/dotnet/roslyn-analyzers)). It caches `ISymbol` lookups for a given `Compilation` to amortize costs across all the analyzers running for a given compilation. It also enforces some best practices when looking for "well known" symbols in the cases of name collisions.

## KnownSymbols

KnownSymbols comes from [typeshape-csharp](https://github.com/eiriktsarpalis/typeshape-csharp) to simplify looking up types in the WellKnownTypeProvider. In this change only types are cached, the member lookups (methods, fields, etc.) are not. However, Roslyn internally caches, so it isn't clear that's needed.

Using the KnownSymbols pattern cleans up the code quite a bit. It does push us away from SyntaxNodes and towards Symbols, however I don't think that's a bad thing because we also want to migrate to the IOperation-based analysis, which already has the corresponding ISymbol available at the start of analysis.

The existing benchmarks show the new and old code (v0.1.0) to be within the margin of error.


MattKotsenas:
Reviewers note: Reviewing by commit makes each refactor clearer. I do however understand this is a larger change, so if there's things to pull out into separate PRs let me know.

rjmurillo (approved):

rjmurillo (approved):

PR #272, opened 2024-11-21
Reduce severity of S3267: Loops should be simplified with "LINQ" expressions

MattKotsenas:
Reduce the severity of S3267: Loops should be simplified with "LINQ" expressions

from a warning to a note / suggestion. This is a stylistic diagnostic that often can suggest refactorings that are (at least subjectively) worse than the code it's replacing. Leaving it as a note though so that you still get the lightbulb so you can see what the refactoring would be in case you like it better.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new diagnostic severity level for code analysis, allowing for more nuanced feedback on loop simplification using LINQ expressions.
	- Added a new package source for local testing, enhancing the development environment.

- **Bug Fixes**
	- Updated severity levels for diagnostic rule S3267 to include both error and note, enhancing clarity in reporting.

- **Documentation**
	- Adjusted documentation for coding conventions to reflect the new severity settings and maintain consistency in formatting rules.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -136,6 +136,7 @@ csharp_style_prefer_range_operator = true:suggestion
 csharp_style_throw_expression = true:suggestion
 csharp_style_unused_value_assignment_preference = discard_variable:suggestion
 csharp_style_unused_value_expression_statement_preference = discard_variable:silent
+dotnet_diagnostic.S3267.severity = suggestion # S3267: Loops should be simplified with "LINQ" expressions
 
 # 'using' directive preferences
 csharp_using_directive_placement = outside_namespace:warning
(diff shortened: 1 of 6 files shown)


MattKotsenas:
SquiggleCop baselines show that this rule appears to be both an `Error` and a `Note`, when I would expect it to only be a `Note`.

I verified the underlying SARIF says the same thing (to rule out a SquiggleCop bug) and filed https://github.com/dotnet/roslyn/issues/76004 to track what appears to be an upstream bug.

MattKotsenas:
Separately, I think .NET 9 introduced a change to how Pack works such that GetPackFromProject is now racing and sometimes failing. Will need to investigate.

rjmurillo (approved):

PR #333, opened 2025-01-30
Update Compiler.props LangVersion

rjmurillo:
The property was tied to a specific (and incidentally the latest version). However, with the release of .NET 9, lang version is no longer 12.0. This change updates the `LangVersion` property to `default`, which is the same as `latestMajor` which is the latest released major version of the compiler

## Summary by CodeRabbit

- **Chores**
	- Updated compiler language version to use default settings instead of a specific version.


rjmurillo:
#331 requires this change

MattKotsenas (approved):

PR #342, opened 2025-02-04
Improve code coverage

rjmurillo:
Addition of new unit tests for extension methods, removing dead code, and fixing some documentation typos

- Added configuration for a new reporting tool to generate coverage reports.
- Enhanced async method analysis and streamlined APIs by removing redundant extension methods.
- Corrected documentation typos to improve clarity.
- Expanded unit test coverage for array operations, enumerable behavior, and asynchronous scenarios.


src/Analyzers/SetupShouldBeUsedOnlyForOverridableMembersAnalyzer.cs
```diff
@@ -108,11 +113,26 @@ private static bool IsTaskResultProperty(IPropertySymbol propertySymbol, MoqKnow
...
 
-        return SymbolEqualityComparer.Default.Equals(propertySymbol.ContainingType, taskOfTType);
+        // Check if the containing type is ValueTask<T>
+        INamedTypeSymbol? valueTaskOfType = knownSymbols.ValueTask1;
+
+        return valueTaskOfType != null &&
+
+               // If ValueTask<T> type cannot be found, we skip it
+               SymbolEqualityComparer.Default.Equals(propertySymbol.ContainingType, valueTaskOfType);
     }
```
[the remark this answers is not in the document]
rjmurillo: Fixed in 8765c5a5

src/Common/EnumerableExtensions.cs
```diff
@@ -10,12 +10,6 @@ internal static class EnumerableExtensions
         return source.DefaultIfNotSingle(static _ => true);
     }
 
-    /// <inheritdoc cref="DefaultIfNotSingle{TSource}(ImmutableArray{TSource}, Func{TSource, bool})"/>
-    public static TSource? DefaultIfNotSingle<TSource>(this ImmutableArray<TSource> source)
```
MattKotsenas, line 14: These are here for symmetry with the `IEnumerable` version. It's OK to remove them if they aren't used, as long as you're also aware that this is highlighting a boxing that we probably should fix elsewhere.

src/Common/WellKnown/KnownSymbols.cs
```diff
@@ -22,21 +22,11 @@ public KnownSymbols(Compilation compilation)
     {
     }
 
-    /// <summary>
-    /// Gets the class <c>System.Threading.Tasks.Task</c>.
-    /// </summary>
-    public INamedTypeSymbol? Task => TypeProvider.GetOrCreateTypeByMetadataName("System.Threading.Tasks.Task");
```
MattKotsenas, line 28: Are you sure you really want to remove this (and the ValueTask version)? I assume this is because they're currently unused, but it feels like you'll just be adding them back soon.

MattKotsenas (approved):

src/Common/WellKnown/KnownSymbols.cs
rjmurillo: I added tests for ValueTask<T>. When I write code that takes Task or ValueTask (without the return type) they always run into the path to exit out of the analyzer using the generic version. If we need them again, we can add them (and we'll have coverage)

src/Common/EnumerableExtensions.cs
rjmurillo: There's only one instance where this is used in production code in AsShouldBeUsedOnlyForInterfaceAnalyzer, which is a LINQ projection and uses IEnumerable<T>

rjmurillo: For the boxing, the ECS analyzer should fire when we have a new boxing issue introduced.

MattKotsenas (approved):

PR #462, opened 2025-06-09
Update main.yml build arguments to reduce flakiness with BuildChecks

rjmurillo:
@coderabbitai

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,7 +48,7 @@ jobs:
       run: dotnet restore
 
     - name: Build
-      run: dotnet build --no-restore --configuration Release /check /bl:./artifacts/logs/release/build.release.binlog
+      run: dotnet build --no-restore --configuration Release /check /p:Deterministic=true /p:UseSharedCompilation=false /p:BuildInParallel=false /nodeResuse:false /bl:./artifacts/logs/release/build.release.binlog
 
     - name: Test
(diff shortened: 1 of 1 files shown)


MattKotsenas:
If we're running into problems with BuildChecks, can we file bugs upstream?

rjmurillo:
> If we're running into problems with BuildChecks, can we file bugs upstream?

@MattKotsenas sure. They produce a bunch of noise [see recent run](https://github.com/rjmurillo/moq.analyzers/actions/runs/15526019694/job/43707147776?pr=458), especially with Nerdbank Git Version, so I'm considering turning them back off. I fixed a couple of errors they surfaced (see #437, #440, #442), but as of now their utility is minimal. I'm considering a separate workflow that just does the build check so the main PR flow isn't jammed up with flaky stuff (and we can dogfood changes there if need be without changing core CI)

rjmurillo:
> If we're running into problems with BuildChecks, can we file bugs upstream?

@MattKotsenas sure. They produce a bunch of noise [see recent run](https://github.com/rjmurillo/moq.analyzers/actions/runs/15526019694/job/43707147776?pr=458), so I'm considering turning them back off. I fixed a couple of errors they surfaced (see #437, #440, #442)

MattKotsenas:
Yeah, I've seen them be noisy too. Just want to make sure upstream is aware, as I'm not yet sure they're ready for adoption?

Issue #1287, opened 2026-07-05
Test harness pins C# 12 (net8.0 ref group); cannot write regression tests for C# 13+ defects


rjmurillo:
## Summary

The analyzer test harness (`AnalyzerVerifier`/`CodeFixVerifier`) compiles embedded test snippets against reference-assembly groups whose highest target is **net8.0**, which pins the C# language version to **C# 12**. This makes it impossible to write regression tests for defects that only manifest under **C# 13+** language features.

## Evidence

While fixing #1241 (`ConstructorArgumentsShouldMatchAnalyzer` throws `InvalidCastException` when a mocked type's constructor uses a **C# 13 params collection**, e.g. `public Foo(params List<int> nums)`), the regression test could not be expressed:

```
// /0/Test1.cs(35,38): error CS0225: The params parameter must be a single dimensional array
```

The shared theory template failed to compile for **every** row because `params List<int>` is not legal in C# 12. The crash itself is real (the cast `((IArrayTypeSymbol)paramsParameter.Type)` fails for non-array params types), but the harness cannot reproduce it.

## Impact

- C# 13 language features (params collections, `field` keyword, partial properties, etc.) cannot be covered by analyzer/code-fix tests.
- Defensive fixes for C# 13-only crash paths ship **without** a proving regression test, weakening the zero-false-positive/zero-crash guarantee.

## Proposed fix

Add a net9.0/net10.0 reference-assembly group (or a per-test `LanguageVersion` override / `CSharpParseOptions` hook) to the verifier so tests can opt into C# 13+ where required. Then backfill the #1241 params-collection regression test.

## Cross-reference

Found while shipping the analyzer crash-hardening batch (#1241, #1242, #1250, #1257).

rjmurillo:
## Triage: the proposed fix is necessary but not sufficient

Investigated while planning the fix. The reference-assembly group (`net8.0`) is not the only thing pinning the language version.

`Directory.Packages.props` pins `Microsoft.CodeAnalysis.CSharp` and `Microsoft.CodeAnalysis.CSharp.Workspaces` to **4.8**. Roslyn 4.8 (VS 17.8) tops out at **C# 12**. The test framework builds the snippet `CSharpCompilation` with that referenced compiler, so:

[...]

Issue #1257, opened 2026-07-02
MockBehavior code fix throws on stale diagnostic shapes; DiagnosticEditProperties accepts undefined EditType values


rjmurillo:
## Summary

An internal code audit found two related robustness holes in the Moq1400/Moq1410 code-fix pipeline (`SetExplicitMockBehaviorFixer`, `SetStrictMockBehaviorFixer` → shared `SetExplicitMockBehaviorCodeAction`):

1. **(a) Stale-shape crash.** `SetExplicitMockBehaviorCodeAction.GetChangedDocumentAsync` calls `SyntaxGeneratorExtensions.InsertArguments`/`ReplaceArgument`, which throw `ArgumentException` (unexpected node kind) or `ArgumentOutOfRangeException` (EditPosition out of range for the node's current argument count). The edit properties travel on the diagnostic and describe the argument-list shape **at analysis time**; if the document changed between analysis and fix application (a stale lightbulb — routine in an IDE), the exception escapes the code action and the IDE surfaces a "code fix encountered an error" failure instead of a graceful no-op.
2. **(b) Undefined enum values round-trip.** `DiagnosticEditProperties.TryGetFromImmutableDictionary` uses `Enum.TryParse`, which accepts *any* numeric string — `"7"` parses successfully to `(EditType)7` — weakening the Try contract. A `(EditType)7` then reaches the code action's `switch`, whose fallback arm throws `InvalidOperationException` at fix-application time.

Both fixes are small, local guards: validate the node shape in `GetChangedDocumentAsync` and return the original document on mismatch; add an `Enum.IsDefined` check to the Try parser.

### (a) `src/CodeFixes/SetExplicitMockBehaviorCodeAction.cs:62-80` — no shape validation before rewriting

```csharp
        SyntaxNode behavior = _behaviorType switch
        {
            BehaviorType.Loose => editor.Generator.MemberAccessExpression(knownSymbols.MockBehaviorLoose),
            BehaviorType.Strict => editor.Generator.MemberAccessExpression(knownSymbols.MockBehaviorStrict),
            _ => throw new InvalidOperationException(),
        };

        SyntaxNode argument = editor.Generator.Argument(behavior);

        SyntaxNode newNode = _editType switch
        {
            DiagnosticEditProperties.EditType.Insert => editor.Generator.InsertArguments(_nodeToFix, _position, argument),
[...]

Issue #1250, opened 2026-07-02
MoqKnownSymbols: `SingleOrDefault` inside cached Lazy permanently poisons MockBehavior field resolution when the compilation defines duplicate members


rjmurillo:
## Summary

An internal code audit found that `MoqKnownSymbols.CreateLazySingleField` uses `SingleOrDefault()` inside a `Lazy<T>` with `LazyThreadSafetyMode.ExecutionAndPublication`. If a compilation contains a source-defined `Moq.MockBehavior` with a duplicated member (a routine mid-edit state — the user is typing, or a merge left two `Default` members), `GetMembers(name)` yields two field symbols, `SingleOrDefault()` throws `InvalidOperationException`, and — because `ExecutionAndPublication` **caches the exception** — every subsequent access to that property rethrows for the lifetime of the compilation. The result is a persistent AD0001 that disables the affected analyzers for the whole session, violating the project's top priority (analyzers must never crash).

## Current behavior

`src/Common/WellKnown/MoqKnownSymbols.cs:641-644`:

```csharp
    private static Lazy<IFieldSymbol?> CreateLazySingleField(INamedTypeSymbol? type, string memberName) =>
        new Lazy<IFieldSymbol?>(
            () => type?.GetMembers(memberName).OfType<IFieldSymbol>().SingleOrDefault(),
            LazyThreadSafetyMode.ExecutionAndPublication);
```

Exactly three properties are built with this helper (`src/Common/WellKnown/MoqKnownSymbols.cs:93-95`):

```csharp
        // MockBehavior enum fields
        _mockBehaviorStrict = CreateLazySingleField(MockBehavior, "Strict");
        _mockBehaviorLoose = CreateLazySingleField(MockBehavior, "Loose");
        _mockBehaviorDefault = CreateLazySingleField(MockBehavior, "Default");
```

exposed as `MockBehaviorStrict` / `MockBehaviorLoose` / `MockBehaviorDefault` (`src/Common/WellKnown/MoqKnownSymbols.cs:278,283,288`).

**Concrete failure input → wrong outcome.** A test project that does not reference the Moq package but declares its own `Moq` namespace (or a broken/duplicated source state while editing) with:

```csharp
namespace Moq
{
    public enum MockBehavior
    {
        Default,
        Default,   // CS0102 — mid-edit duplicate; Roslyn still creates BOTH field symbols
        Strict,
        Loose,
    }

    public class Mock<T> where T : class
    {
        public Mock() { }
[...]

PR #1288, opened 2026-07-05
fix(analyzers): harden four crash/poison paths (#1241, #1242, #1250, #1257)

### LINKED ISSUE #1241: ConstructorArgumentsShouldMatchAnalyzer crashes (InvalidCastException / AD0001) on C# 13 params collections
## Summary

An internal code audit found that `ConstructorArgumentsShouldMatchAnalyzer` (rules **Moq1001**/**Moq1002**) performs an unguarded cast `(IArrayTypeSymbol)paramsParameter.Type` when validating `params` constructor arguments. On modern Roslyn hosts (VS 17.10+/.NET 9+ SDK compiling C# 13), `IParameterSymbol.IsParams` is also `true` for C# 13 *params collections* (`params ReadOnlySpan<int>`, `params IEnumerable<int>`, `params List<int>`, ...), whose `Type` is **not** an `IArrayTypeSymbol`. The cast throws `InvalidCastException`, Roslyn reports **AD0001**, and the analyzer is disabled for the rest of the session. **Severity: Critical** (analyzer crash, violates the project's #1 priority of never crashing the host compiler).

## Current behavior

`src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs:376-396` (cast at line 383):

```csharp
    /// <summary>
    /// Verifies that all arguments in the params position are convertible to the params element type.
    /// </summary>
    private static bool AllParamsArgumentsMatch(
        IMethodSymbol constructor,
        FilteredArgumentList arguments,
        int fixedParametersCount,
        SyntaxNodeAnalysisContext context)
    {
        IParameterSymbol paramsParameter = constructor.Parameters[^1];
        ITypeSymbol paramsElementType = ((IArrayTypeSymbol)paramsParameter.Type).ElementType;

        for (int parameterIndex = fixedParametersCount; parameterIndex < arguments.Count; parameterIndex++)
        {
            Conversion conversionType = context.SemanticModel.ClassifyConversion(arguments[parameterIndex].Expression, paramsElementType);

            if (!conversionType.Exists)
            {
                return false;
            }
        }

        return true;
    }
```

The method is entered from `IsConstructorMatch` (`src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs:289-322`):

```csharp
        bool hasParams = constructor.Parameters.Length > 0 && constructor.Parameters[^1].IsParams;   // line 294
        ...
        if (hasParams && !AllParamsArgumentsMatch(constructor, arguments, fixedParametersCount, context))   // line 316
        {
            return false;
        }
```

**Failure scenario.** The analyzer ships as netstandard2.0 pinned to the Roslyn 4.8 *API*, but it executes inside whatever Roslyn the consumer's IDE/SDK hosts. Given user code compiled as C# 13:

```csharp
public class Foo
{
    public Foo(params System.ReadOnlySpan<int> xs) { }
}

var mock = new Mock<Foo>(1, 2, 3);
```

On a host with Roslyn 4.10+, `constructor.Parameters[^1].IsParams == true` (Roslyn sets `IsParams` for both params arrays and params collections), so `hasParams` is true, argument-count and fixed-parameter checks pass, `AllParamsArgumentsMatch` is entered, and `(IArrayTypeSymbol)paramsParameter.Type` throws `InvalidCastException` on the `ReadOnlySpan<int>` named type. Roslyn converts this into an AD0001 diagnostic and disables the analyzer for the session.

**Downstream trace — what the guard's return value means.** Callers of `AllParamsArgumentsMatch`:

1. `IsConstructorMatch` line 316: `false` ⇒ this constructor is treated as *not matching*.
2. `AnyConstructorsFound` (lines 257-279): if no constructor matches, returns `false`.
3. `VerifyClassMockAttempt` (lines 457-488, report at 482-487): a `false` result reports **Moq1002** ("Could not find a matching constructor").

Therefore the originally sketched guard `if (... is not IArrayTypeSymbol) return false;` would convert the crash into a **false positive**: `new Mock<Foo>(1, 2, 3)` against `Foo(params ReadOnlySpan<int>)` — code the C# compiler accepts — would be flagged Moq1002 whenever the params-collection constructor is the only candidate. The correct conservative bail is to **return `true`** ("cannot verify the variadic tail with the Roslyn 4.8 API; assume it matches"), which produces **no diagnostic**. A false negative on exotic params-collection mismatches is acceptable; a crash or a false positive is not. This mirrors the analyzer's existing convention of suppressing when it cannot verify (see `AnyConstructorsFound` returning `null` for the parenthesized-lambda case, lines 270-276).

## Desired behavior

- The analyzer never throws when a candidate constructor's last parameter is a C# 13 params collection.
- Params-**array** constructors (`params T[]`) keep exactly the current match/mismatch behavior.
- Params-**collection** constructors conservatively count as matching (no Moq1001/Moq1002 reported) because the pinned Roslyn 4.8 API cannot inspect them (`IParameterSymbol.IsParamsCollection` does not exist in 4.8).

## Implementation plan

1. Edit `src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs`, replacing the whole `AllParamsArgumentsMatch` method (currently lines 373-396, including its XML doc) with:

```csharp
    /// <summary>
    /// Verifies that all arguments in the params position are convertible to the params element type.
    /// </summary>
    private static bool AllParamsArgumentsMatch(
        IMethodSymbol constructor,
        FilteredArgumentList arguments,
        int fixedParametersCount,
        SyntaxNodeAnalysisContext context)
    {
        IParameterSymbol paramsParameter = constructor.Parameters[^1];

        // On modern Roslyn hosts, IsParams is also true for C# 13 params collections
        // (e.g. `params ReadOnlySpan<int>`), whose type is not an array. The pinned
        // Roslyn 4.8 API cannot inspect params collections (IsParamsCollection does
        // not exist), so conservatively treat the arguments as matching: returning
        // false here would report a false Moq1002 on code the compiler accepts, and
        // casting unconditionally crashes the analyzer (AD0001).
        if (paramsParameter.Type is not IArrayTypeSymbol paramsArrayType)
        {
            return true;
        }

        ITypeSymbol paramsElementType = paramsArrayType.ElementType;

        for (int parameterIndex = fixedParametersCount; parameterIndex < arguments.Count; parameterIndex++)
        {
            Conversion conversionType = context.SemanticModel.ClassifyConversion(arguments[parameterIndex].Expression, paramsElementType);

            if (!conversionType.Exists)
            {
                return false;
            }
        }

        return true;
    }
```

   No other member changes. No using-directive changes (`IArrayTypeSymbol`, `IParameterSymbol`, `ITypeSymbol` already resolve via the project's global usings, as evidenced by the current code in this file).

2. No documentation change is required: on the toolchains the repo can test (Roslyn 4.8 API surface), observable behavior of Moq1001/Moq1002 is unchanged — this is a pure crash guard. Do **not** touch `AnalyzerReleases.Shipped.md` or `AnalyzerReleases.Unshipped.md` (no severity/category change).

## Tests

**Honest limitation — the C# 13 crash scenario is not testable in this suite.** The test project compiles test sources with the pinned Roslyn 4.8 compiler (`Microsoft.CodeAnalysis.CSharp.Workspaces` 4.8 via `Directory.Packages.props`; `Microsoft.CodeAnalysis.CSharp.CodeFix.Testing` uses `LanguageVersion.Default`, which for Roslyn 4.8 is C# 12). Roslyn 4.8:

- cannot parse `params ReadOnlySpan<int>` in source (params collections require Roslyn 4.10+/C# 13; 4.8's `LanguageVersion.Preview` does not include them), and
- does not map metadata `ParamCollectionAttribute` to `IsParams`, so a metadata-reference route cannot reproduce `IsParams == true` with a non-array type either.

The guarded branch is therefore *scenario-untestable* with the pinned toolchain; the guard is defense-in-depth for newer hosts. State this in a code comment (already included in the replacement above) and in the PR description. Do **not** invent a fake `IParameterSymbol`; the repo does not mock Roslyn symbol interfaces.

**What must be tested instead** — lock the params-*array* behavior of the edited method so the refactor cannot regress it:

1. All existing rows in `tests/Moq.Analyzers.Test/ConstructorArgumentsShouldMatchAnalyzerTests.cs` `TestData()` must keep passing, in particular the `ClassWithParams` rows (lines 12-18) which exercise the `true` path of `AllParamsArgumentsMatch`. The fixture (same file, lines 120-124) is:

```csharp
                internal class ClassWithParams
                {
                    public ClassWithParams(params DateTime[] dates) { }
                    public ClassWithParams(string s, params DateTime[] dates) { }
                }
```

2. Add rows to `TestData()` in `tests/Moq.Analyzers.Test/ConstructorArgumentsShouldMatchAnalyzerTests.cs` that exercise the `false` path (mismatched params-array tail) — insert directly after the existing `ClassWithParams` rows (after line 18):

```csharp
            // params-array tail with a non-convertible argument must still be flagged
            // (locks AllParamsArgumentsMatch's mismatch path after the params-collection guard)
            ["""new Mock<ClassWithParams>{|Moq1002:(DateTime.Now, 42)|};"""],
            ["""new Mock<ClassWithParams>{|Moq1002:(MockBehavior.Default, DateTime.Now, 42)|};"""],
```

   Rationale: `(DateTime.Now, 42)` matches neither `ClassWithParams(params DateTime[])` (the `42` tail argument has no conversion to `DateTime`, so `AllParamsArgumentsMatch` returns `false`) nor `ClassWithParams(string, params DateTime[])` (`DateTime.Now` has no conversion to `string`). These rows are multiplied automatically across 2 namespaces × 2 Moq versions (4.8.2, 4.18.4) by `.WithNamespaces().WithMoqReferenceAssemblyGroups()` at the end of `TestData()`; params-array constructor matching is not Moq-version-sensitive, so both versions are fine.

3. No new test file is needed; the test method `ShouldAnalyzeConstructorArguments` ([Theory] + [MemberData(nameof(TestData))]) already consumes the data.

## Constraints

- Analyzer assemblies target netstandard2.0; Roslyn SDK pinned to Microsoft.CodeAnalysis **4.8** (ADR-003). Do NOT use `IParameterSymbol.IsParamsCollection` — it does not exist in 4.8; the fix must use the type-pattern check shown above.
- No new package dependencies (ADR-004/ADR-005).
- Symbol-based detection per ADR-001; `MoqKnownSymbols` created once per compilation per ADR-006 (unchanged here); ADR-007 IOperation migration is out of scope for this fix.
- Tests live in `tests/Moq.Analyzers.Test/` and follow repo patterns: `{|Moq1002:...|}` markup asserting ID + exact span; `[Theory]` + `[MemberData]` with `public static IEnumerable<object[]>`; rows multiplied across 2 namespaces × 2 Moq versions via `TestDataExtensions`.
- Do not touch `AnalyzerReleases.Shipped.md`; no `AnalyzerReleases.Unshipped.md` change (pure bugfix, no severity/category change).
- Do not claim test coverage of the C# 13 path — it is not achievable with the pinned test compiler (see Tests section).

## Validation

```bash
dotnet format --verify-no-changes   # or `dotnet format` before committing
dotnet build /p:PedanticMode=true   # must be 0 warnings, 0 errors
dotnet test --settings ./build/targets/tests/test.runsettings
```

## Acceptance criteria

- [ ] `AllParamsArgumentsMatch` contains no unguarded cast to `IArrayTypeSymbol`; a non-array `params` parameter type returns `true` (conservative: no diagnostic) with the explanatory comment.
- [ ] Existing `ClassWithParams` positive rows (no diagnostic) still pass.
- [ ] New `{|Moq1002:(DateTime.Now, 42)|}` mismatch rows pass on both Moq versions and both namespace variants.
- [ ] Full test suite green; build has 0 warnings including `/p:PedanticMode=true`.
- [ ] No changes to `AnalyzerReleases.*.md`, no new dependencies, no APIs beyond Roslyn 4.8.

## Out of scope

- Actually validating conversions against params-collection element types (requires Roslyn ≥ 4.10 APIs; blocked by ADR-003 until the minimum Roslyn is raised).
- Migrating this analyzer to `RegisterOperationAction` (ADR-007) — tracked separately.
- The target-typed `new` false negative in the same analyzer (separate issue).
- Hot-path allocation cleanup in `VerifyClassMockAttempt` (audit item A-13, separate).

rjmurillo:
Fixes #1241. Also addresses #1242, #1250, #1257.

## Problem
`ConstructorArgumentsShouldMatchAnalyzer` (Moq1002) had several correctness and robustness defects:

1. **Crash (#1241)**: a `params` parameter whose type is not a plain array (a C# 13 params collection such as `params ReadOnlySpan<T>`, `params IEnumerable<T>`) caused an unhandled cast/exception because the code assumed `IArrayTypeSymbol`.
2. Related edge cases (#1242, #1250, #1257) around argument matching against `params` tails.

## Fix
Rewrote the `params` tail handling in `AllParamsArgumentsMatch`:

- **Normal-form call**: a single trailing argument that converts to the whole params type (e.g. passing a `List<int>` for `params List<int>`) is accepted directly.
- **Expanded-form call**: each trailing argument is validated against the params **element type**, derived by `GetParamsElementType`:
  - array → element type,
  - `Span<T>`/`ReadOnlySpan<T>` (matched by symbol identity via `IsSpanOrReadOnlySpan`, restricted to the top-level `System` types) → `T`,
  - `IEnumerable<T>` (or a type implementing it) → `T`.
- **Unmodeled exotic collection shapes** → fail open (skip validation). This preserves the project's zero-false-positive priority: we never emit Moq1002 from a guessed element type.

This removes the crash and prevents both false positives (normal-form and exotic shapes) and false negatives (expanded-form element mismatches remain detected).

## Tests
Constructor-arguments analyzer suite: 400 tests passing. Note: positive tests that need C# 13 `params` collection *syntax* to compile are blocked by the test harness pinning C# 12 (tracked in #1287); the full CI matrix compiles and exercises the newer language version as a backstop.

## Validation
- `dotnet build`: 0 warnings, 0 errors.
- Full test suite: 3358 passed, 0 failed.
- Constructor-arguments targeted: 400 passed.
- Moq versions: analyzer logic is symbol-based and behaves identically for Moq 4.8.2 and 4.18.4.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved constructor `params` matching by deriving the element type for supported `params` forms and handling unmodelable shapes safely.
  * Tightened diagnostic edit-type parsing to accept only the defined edit kinds.
  * Prevented issues when analyzing setup invocations that have no arguments.
  * Made well-known field resolution deterministic when multiple matches exist, avoiding exceptions.
* **Tests**
  * Added coverage to ensure undefined numeric edit-type values are rejected.

### DIFF
--- a/src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs
+++ b/src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs
@@ -391,6 +391,17 @@ public class ConstructorArgumentsShouldMatchAnalyzer : DiagnosticAnalyzer
             return true;
         }
 
+        // For non-array params collections, a single trailing argument may be the collection
+        // itself passed in normal form (for example `new List<int>()` for `params List<int>`)
+        // rather than an expanded element. Accept it when it converts to the params parameter
+        // type before validating expanded elements. Arrays keep the existing expanded-only path.
+        if (paramsParameter.Type is not IArrayTypeSymbol
+            && arguments.Count - fixedParametersCount == 1
+            && context.SemanticModel.ClassifyConversion(arguments[fixedParametersCount].Expression, paramsParameter.Type).Exists)
+        {
+            return true;
+        }
+
         for (int parameterIndex = fixedParametersCount; parameterIndex < arguments.Count; parameterIndex++)
         {
             Conversion conversionType = context.SemanticModel.ClassifyConversion(arguments[parameterIndex].Expression, paramsElementType);
@@ -440,14 +451,31 @@ public class ConstructorArgumentsShouldMatchAnalyzer : DiagnosticAnalyzer
             }
         }
 
-        // ReadOnlySpan<T> and Span<T> are ref structs that do not implement IEnumerable<T>,
-        // so fall back to the single generic type argument.
-        if (namedType is { IsGenericType: true, TypeArguments.Length: 1 })
+        // ReadOnlySpan<T> and Span<T> are ref structs that do not implement IEnumerable<T>.
+        // Match them by symbol identity rather than a broad "single generic argument" heuristic,
+        // so unmodeled shapes fall through and are skipped rather than mis-validated.
+        if (IsSpanOrReadOnlySpan(namedType))
         {
             return namedType.TypeArguments[0];
         }
 
         return null;
+    }
+
+    /// <summary>
+    /// Determines whether a type is <see cref="System.Span{T}" /> or
+    /// <see cref="System.ReadOnlySpan{T}" /> by symbol identity.
+    /// </summary>
+    private static bool IsSpanOrReadOnlySpan(INamedTypeSymbol namedType)
+    {
+        INamedTypeSymbol definition = namedType.OriginalDefinition;
+
+        if (definition.ContainingNamespace is not { Name: "System", ContainingNamespace.IsGlobalNamespace: true })
+        {
+            return false;
+        }
+
+        return definition.MetadataName is "Span`1" or "ReadOnlySpan`1";
     }
 
     private static (bool IsEmpty, Location Location) ConstructorIsEmpty(
--- a/src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs
+++ b/src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs
@@ -470,7 +470,8 @@ public class ConstructorArgumentsShouldMatchAnalyzer : DiagnosticAnalyzer
     {
         INamedTypeSymbol definition = namedType.OriginalDefinition;
 
-        if (definition.ContainingNamespace is not { Name: "System", ContainingNamespace.IsGlobalNamespace: true })
+        if (definition.ContainingType is not null
+            || definition.ContainingNamespace is not { Name: "System", ContainingNamespace.IsGlobalNamespace: true })
         {
             return false;
         }
--- a/src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs
+++ b/src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs
@@ -380,7 +380,17 @@ public class ConstructorArgumentsShouldMatchAnalyzer : DiagnosticAnalyzer
         SyntaxNodeAnalysisContext context)
     {
         IParameterSymbol paramsParameter = constructor.Parameters[^1];
-        ITypeSymbol paramsElementType = ((IArrayTypeSymbol)paramsParameter.Type).ElementType;
+
+        // A `params` parameter is normally an array, but C# 13 allows params collections
+        // (for example `params ReadOnlySpan<T>` or `params List<T>`) whose type is not an
+        // array. We cannot reliably classify element conversions for those shapes here, so
+        // skip params validation rather than crash on an invalid cast.
+        if (paramsParameter.Type is not IArrayTypeSymbol paramsArrayType)
+        {
+            return true;
+        }
+
+        ITypeSymbol paramsElementType = paramsArrayType.ElementType;
 
         for (int parameterIndex = fixedParametersCount; parameterIndex < arguments.Count; parameterIndex++)
         {
--- a/src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs
+++ b/src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs
@@ -383,14 +383,13 @@ public class ConstructorArgumentsShouldMatchAnalyzer : DiagnosticAnalyzer
 
         // A `params` parameter is normally an array, but C# 13 allows params collections
         // (for example `params ReadOnlySpan<T>` or `params List<T>`) whose type is not an
-        // array. We cannot reliably classify element conversions for those shapes here, so
-        // skip params validation rather than crash on an invalid cast.
-        if (paramsParameter.Type is not IArrayTypeSymbol paramsArrayType)
+        // array. Derive the element type for the known collection shapes so trailing
+        // arguments are still validated; only skip when the shape is genuinely unmodelable.
+        ITypeSymbol? paramsElementType = GetParamsElementType(paramsParameter.Type);
+        if (paramsElementType is null)
         {
             return true;
         }
-
-        ITypeSymbol paramsElementType = paramsArrayType.ElementType;
 
         for (int parameterIndex = fixedParametersCount; parameterIndex < arguments.Count; parameterIndex++)
         {
@@ -403,6 +402,52 @@ public class ConstructorArgumentsShouldMatchAnalyzer : DiagnosticAnalyzer
         }
 
         return true;
+    }
+
+    /// <summary>
+    /// Determines the element type for a <see langword="params" /> parameter.
+    /// </summary>
+    /// <param name="paramsType">The declared type of the <see langword="params" /> parameter.</param>
+    /// <returns>
+    /// The element type for arrays and known params-collection shapes (C# 13
+    /// <c>ReadOnlySpan&lt;T&gt;</c>, <c>Span&lt;T&gt;</c>, and any type implementing
+    /// <see cref="System.Collections.Generic.IEnumerable{T}" />); otherwise
+    /// <see langword="null" /> when the shape cannot be modeled.
+    /// </returns>
+    private static ITypeSymbol? GetParamsElementType(ITypeSymbol paramsType)
+    {
+        if (paramsType is IArrayTypeSymbol arrayType)
+        {
+            return arrayType.ElementType;
+        }
+
+        if (paramsType is not INamedTypeSymbol namedType)
+        {
+            return null;
+        }
+
+        // Types that are, or implement, IEnumerable<T> (for example List<T>).
+        if (namedType.OriginalDefinition.SpecialType == SpecialType.System_Collections_Generic_IEnumerable_T)
+        {
+            return namedType.TypeArguments[0];
+        }
+
+        foreach (INamedTypeSymbol interfaceType in namedType.AllInterfaces)
+        {
+            if (interfaceType.OriginalDefinition.SpecialType == SpecialType.System_Collections_Generic_IEnumerable_T)
+            {
+                return interfaceType.TypeArguments[0];
+            }
+        }
+
+        // ReadOnlySpan<T> and Span<T> are ref structs that do not implement IEnumerable<T>,
+        // so fall back to the single generic type argument.
+        if (namedType is { IsGenericType: true, TypeArguments.Length: 1 })
+        {
+            return namedType.TypeArguments[0];
+        }
+
+        return null;
     }
 
     private static (bool IsEmpty, Location Location) ConstructorIsEmpty(
--- a/src/Common/DiagnosticEditProperties.cs
+++ b/src/Common/DiagnosticEditProperties.cs
@@ -65,7 +65,10 @@ internal record DiagnosticEditProperties
             return false;
         }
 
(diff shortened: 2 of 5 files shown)


rjmurillo:
## Review feedback addressed

**`ConstructorArgumentsShouldMatchAnalyzer` — non-array `params` collections (CodeRabbit CHANGES_REQUESTED, Codex P2)** — commit f1c0f66 + normal-form follow-up
The non-array `params` guard no longer blanket-accepts trailing arguments. `GetParamsElementType` now derives the element type for arrays, `IEnumerable<T>` collections (e.g. `List<T>`), and `Span<T>`/`ReadOnlySpan<T>`, then validates expanded arguments against it. A single trailing argument that converts to the whole collection type (normal form) is accepted. It only bails (returns `true`) when the shape is genuinely unmodelable. Array behavior is byte-for-byte unchanged. A harness positive test for C# 13 params collections is blocked by #1287 (test harness pins C# 12).

**`SemanticModelExtensions` — nullable `SeparatedSyntaxList` wrapper (Gemini medium, CodeRabbit)** — commit 32c76ed
Switched to the idiomatic property pattern `setupMethodInvocation?.ArgumentList is not { Arguments: { Count: > 0 } arguments }`, removing the nullable-struct `.Value` access.

**`MoqKnownSymbols.CreateLazySingleField` — `FirstOrDefault` on duplicates (Copilot)** — commit 32c76ed
Switched to the codebase's `DefaultIfNotSingle` helper so ambiguous duplicate fields resolve to `null` (skip) deterministically instead of picking an arbitrary match. Remarks updated.

An adversarial GPT-5.5 re-review of the full diff after these changes returned **CLEAN**. Full suite: **3358 passed, 0 failed, 0 warnings**.

rjmurillo:
## Review round 2 addressed — commit d6b0241

**Normal-form params collections (Codex P2)** — the normal-form acceptance was validated locally but not included in the earlier push; it is now committed. Before expanded-element validation, a single trailing argument that converts to the params collection type is accepted, so `new Mock<C>(new List<int>())` for `C(params List<int>)` no longer reports Moq1002. Arrays keep the expanded-only path.

**Span detection heuristic too broad (CodeRabbit)** — replaced the "any single-generic-argument type" fallback with symbol-identity detection (`IsSpanOrReadOnlySpan`, matching `System.Span\`1`/`System.ReadOnlySpan\`1`), consistent with the codebase's symbol-based mandate.

**Unmodeled custom collections (Codex P2 "reject unmodeled")** — deliberately **not** deriving iteration types for arbitrary `[CollectionBuilder]`/`Add(T)` shapes. This repo weights zero false positives above catching every case: inferring an element type for exotic shapes risks diverging from the compiler's params-collection semantics and producing false Moq1002 (a concern Copilot also raised). Unmodeled shapes fail-open (skip validation), which can only ever miss a diagnostic, never emit a wrong one, and never crash. This is the safe, intentional tradeoff for a rare C# 13 shape.

**Copilot (behavior change on C# 13 hosts)** — acknowledged. The change validates only the common, well-defined shapes (arrays unchanged, `IEnumerable<T>`, `Span<T>`/`ReadOnlySpan<T>`) and fails open elsewhere, so it cannot introduce false positives on shapes it does not model. A harness positive test is blocked by #1287 (harness pins C# 12).

Build: 0 warnings. Constructor analyzer tests: **400 passed**.

rjmurillo:
## Adversarial review round 3 — commit c5954c3

**Nested `System.Foo.Span\<T\>` false match (GPT-5.5 Medium)** — fixed. `IsSpanOrReadOnlySpan` now requires `definition.ContainingType is null`, so only the top-level `System.Span\<T\>`/`System.ReadOnlySpan\<T\>` match. A nested type with those metadata names in namespace `System` no longer qualifies.

Build: 0 warnings. Constructor-arguments tests: 400 passed, 0 failed.

PR #1289, opened 2026-07-05
fix(analyzers): Moq1300 false positive on open type parameters / unresolved types (#1251)

### LINKED ISSUE #1251: Moq1300 fires at Error severity on unresolved (error) and generic type-parameter As<T> arguments; test coverage is thin (4 rows)
## Summary

An internal code audit found that `AsShouldBeUsedOnlyForInterfaceAnalyzer` (Moq1300, **Error** severity) reports on any `As<T>()` type argument whose `TypeKind` is not `Interface`. Two `TypeKind`s that satisfy that condition should never be flagged:

- `TypeKind.Error` — mid-edit code (`mock.As<INonExistent>()` while the interface is being typed or the using is missing) gets a red-squiggle **error** diagnostic on code that will be fine once it resolves.
- `TypeKind.TypeParameter` — `m.As<T>()` inside a generic forwarder (`where T : class`) is valid whenever `T` is instantiated with an interface; flagging the open generic is a false positive.

Both cases were verified against the pinned Roslyn (4.8.0): the invocation binds to an `IInvocationOperation` whose `TargetMethod.TypeArguments[0]` has `TypeKind.Error` / `TypeKind.TypeParameter` respectively, so the analyzer reports today.

The audit's test-quality pass also ranked Moq1300 as the weakest-covered analyzer: `tests/Moq.Analyzers.Test/AsAcceptOnlyInterfaceAnalyzerTests.cs` has only 4 data rows — no generic interfaces, no nested generics, no chained `As<I1>().As<I2>()`, no `As` on a mock variable, no delegate/struct/enum arguments, no malformed code.

## Current behavior

`src/Analyzers/AsShouldBeUsedOnlyForInterfaceAnalyzer.cs:64-98` (the full analysis callback):

```csharp
    private static void Analyze(OperationAnalysisContext context, ImmutableArray<IMethodSymbol> wellKnownAsMethods)
    {
        // This should always be an invocation operation since we registered for OperationKind.Invocation
        Debug.Assert(context.Operation is IInvocationOperation, "Expected IInvocationOperation");

        if (context.Operation is not IInvocationOperation invocationOperation)
        {
            return;
        }

        IMethodSymbol targetMethod = invocationOperation.TargetMethod;
        if (!targetMethod.IsInstanceOf(wellKnownAsMethods))
        {
            return;
        }

        ImmutableArray<ITypeSymbol> typeArguments = targetMethod.TypeArguments;
        if (typeArguments.Length != 1)
        {
            return;
        }

        if (typeArguments[0] is ITypeSymbol { TypeKind: not TypeKind.Interface } typeSymbol)
        {
            // Find the first As<T> generic type argument and report the diagnostic on it
            GenericNameSyntax? asGeneric = invocationOperation.Syntax
                .DescendantNodes()
                .OfType<GenericNameSyntax>()
                .FirstOrDefault(x => string.Equals(x.Identifier.ValueText, "As", StringComparison.Ordinal));

            TypeSyntax? typeArg = asGeneric?.TypeArgumentList.Arguments.FirstOrDefault();
            Location location = typeArg?.GetLocation() ?? invocationOperation.Syntax.GetLocation();
            context.ReportDiagnostic(location.CreateDiagnostic(Rule, typeSymbol.ToDisplayString()));
        }
    }
```

The guard at line 86 is `TypeKind: not TypeKind.Interface`, so:

**Failure input 1 (error type):**

```csharp
var mock = new Mock<SampleClass>();
mock.As<INonExistent>();   // CS0246 on INonExistent
```

Roslyn still binds the invocation (verified: `IInvocationOperation`, type argument `TypeKind.Error`), so Moq1300 is reported at **Error** severity on top of the compiler's own CS0246 — noisy while typing (`mock.As<IServ` mid-keystroke resolves the same way once the argument list closes). Contrast `src/Analyzers/ConstructorArgumentsShouldMatchAnalyzer.cs:414`, which explicitly guards `if (mockedClass is IErrorTypeSymbol)` for exactly this reason.

**Failure input 2 (type parameter):**

```csharp
internal static class MockExtensions
{
    public static Mock<TInterface> AsInterface<TInterface>(Mock<SampleClass> mock)
        where TInterface : class
    {
        return mock.As<TInterface>();   // Moq1300 today — but valid for AsInterface<ISampleInterface>(...)
    }
}
```

Verified: the type argument surfaces as `TypeKind.TypeParameter` and is flagged.

**Additional verified facts** (relevant to tests below):

- Delegate type arguments (`As<System.Action>()`) satisfy Moq's `where TInterface : class` constraint, bind normally, and have `TypeKind.Delegate` → flagged today. This is **correct** (Moq throws `ArgumentException` at runtime for non-interface `As`) and must keep flagging.
- Struct/enum type arguments violate the `class` constraint (CS0452). Verified on the pinned Roslyn: the bound node is an `IInvalidOperation`, **not** an `IInvocationOperation`, so the analyzer never sees them and no Moq1300 is reported — the compiler error is the user's signal. Tests should pin this as a no-crash negative.

## Desired behavior

Skip `TypeKind.Error` and `TypeKind.TypeParameter` before reporting; everything else keeps current behavior (classes, abstract classes, delegates, etc. are flagged).

**Type-parameter policy (decision + justification):** skip type parameters entirely, without inspecting constraints. A constraint-based refinement (flag only when a class-type constraint provably makes `T` a non-interface, e.g. `where T : SampleClass`) would catch a rare misuse at the cost of non-trivial constraint analysis; the simple skip keeps the rule crash-safe and FP-free at Error severity. Accepted trade-off: a false negative for generic forwarders whose constraint already proves non-interface — the runtime `ArgumentException` and the rule firing at every concrete instantiation site still cover that case in practice.

## Implementation plan

1. In `src/Analyzers/AsShouldBeUsedOnlyForInterfaceAnalyzer.cs`, replace the `Analyze` method (currently lines 64-98) with:

```csharp
    private static void Analyze(OperationAnalysisContext context, ImmutableArray<IMethodSymbol> wellKnownAsMethods)
    {
        // This should always be an invocation operation since we registered for OperationKind.Invocation
        Debug.Assert(context.Operation is IInvocationOperation, "Expected IInvocationOperation");

        if (context.Operation is not IInvocationOperation invocationOperation)
        {
            return;
        }

        IMethodSymbol targetMethod = invocationOperation.TargetMethod;
        if (!targetMethod.IsInstanceOf(wellKnownAsMethods))
        {
            return;
        }

        ImmutableArray<ITypeSymbol> typeArguments = targetMethod.TypeArguments;
        if (typeArguments.Length != 1)
        {
            return;
        }

        ITypeSymbol typeSymbol = typeArguments[0];

        // Do not report on unresolved types (mid-edit code; the compiler already reports CS0246)
        // or on open type parameters (a generic forwarder's As<T>() is valid whenever T is
        // instantiated with an interface). Cf. the IErrorTypeSymbol guard in
        // ConstructorArgumentsShouldMatchAnalyzer.
        if (typeSymbol.TypeKind is TypeKind.Error or TypeKind.TypeParameter)
        {
            return;
        }

        if (typeSymbol.TypeKind is not TypeKind.Interface)
        {
            // Find the first As<T> generic type argument and report the diagnostic on it
            GenericNameSyntax? asGeneric = invocationOperation.Syntax
                .DescendantNodes()
                .OfType<GenericNameSyntax>()
                .FirstOrDefault(x => string.Equals(x.Identifier.ValueText, "As", StringComparison.Ordinal));

            TypeSyntax? typeArg = asGeneric?.TypeArgumentList.Arguments.FirstOrDefault();
            Location location = typeArg?.GetLocation() ?? invocationOperation.Syntax.GetLocation();
            context.ReportDiagnostic(location.CreateDiagnostic(Rule, typeSymbol.ToDisplayString()));
        }
    }
```

   No other member of the file changes. No new usings (the file already has `System.Diagnostics` and `Microsoft.CodeAnalysis.Operations`).

2. Update `docs/rules/Moq1300.md`: add a short "Notes" paragraph stating the rule does not report on unresolved type arguments (the compiler reports those) or on open generic type parameters (`As<T>()` in a generic method), and continues to report on any concrete non-interface type including classes and delegates.

3. Do not touch `src/Analyzers/AnalyzerReleases.Shipped.md`.

## Tests

All in `tests/Moq.Analyzers.Test/AsAcceptOnlyInterfaceAnalyzerTests.cs`. The existing `Template` (lines 29-56) declares `ISampleInterface`, `BaseSampleClass`, `SampleClass`, and a `UnitTest.Test()` wrapper.

### 1. Extend the template

Add to the template's declarations (after `SampleClass`):

```csharp
              public interface IOtherInterface
              {
                  void Do();
              }

              public interface IGenericInterface<T>
              {
                  T Get();
              }
```

### 2. Extend `TestData()` (currently 4 rows) with the coverage gaps

```csharp
    public static IEnumerable<object[]> TestData()
    {
        return new object[][]
        {
            ["""new Mock<BaseSampleClass>().As<{|Moq1300:BaseSampleClass|}>();"""],
            ["""new Mock<BaseSampleClass>().As<{|Moq1300:SampleClass|}>();"""],
            ["""new Mock<SampleClass>().As<ISampleInterface>();"""],
            ["""new Mock<SampleClass>().As<ISampleInterface>().Setup(x => x.Calculate(It.IsAny<int>(), It.IsAny<int>())).Returns(10);"""],

            // Generic interface type arguments
            ["""new Mock<SampleClass>().As<IGenericInterface<int>>();"""],
            ["""new Mock<SampleClass>().As<System.Collections.Generic.IList<int>>();"""],

            // Nested generic interface type argument
            ["""new Mock<SampleClass>().As<IGenericInterface<System.Collections.Generic.List<int>>>();"""],

            // Chained As calls: both interfaces - no diagnostics
            ["""new Mock<SampleClass>().As<ISampleInterface>().As<IOtherInterface>();"""],

            // Chained As calls: second one is a class - flag only the second
            ["""new Mock<SampleClass>().As<ISampleInterface>().As<{|Moq1300:SampleClass|}>();"""],

            // As on a mock variable (not inline on the creation)
            ["""var mock = new Mock<SampleClass>(); mock.As<ISampleInterface>();"""],
            ["""var mock = new Mock<SampleClass>(); mock.As<{|Moq1300:SampleClass|}>();"""],

            // Delegate type argument: satisfies Moq's `class` constraint, is not an
            // interface (TypeKind.Delegate), and fails at runtime - must keep flagging.
            ["""new Mock<SampleClass>().As<{|Moq1300:System.Action|}>();"""],
        }.WithNamespaces().WithMoqReferenceAssemblyGroups();
    }
```

Note on the chained rows: verify the reported span lands on the second `As`'s type argument. The analyzer locates the **first** `GenericNameSyntax` named `As` among the invocation's descendant nodes; for the outer invocation `....As<ISampleInterface>().As<SampleClass>()` the receiver's `As<ISampleInterface>` is also a descendant, so if the markup fails, the observed span will reveal the (pre-existing) location-picking quirk — in that case pin the observed span with a `// location quirk: reported on the first As<T> in the chain` comment rather than changing the location logic in this issue.

### 3. New test — type-parameter negative

```csharp
    public static IEnumerable<object[]> TypeParameterTestData()
    {
        return new object[][]
        {
            // As<T>() in a generic forwarder is valid whenever T is instantiated
            // with an interface; the open generic must not be flagged.
            [string.Empty],
        }.WithNamespaces().WithMoqReferenceAssemblyGroups();
    }

    [Theory]
    [MemberData(nameof(TypeParameterTestData))]
    public async Task ShouldNotReportWhenAsTypeArgumentIsTypeParameter(string referenceAssemblyGroup, string @namespace, string _)
    {
        await Verifier.VerifyAnalyzerAsync(
            $$"""
              {{@namespace}}

              public interface ISampleInterface
              {
                  int Calculate(int a, int b);
              }

              public class SampleClass
              {
                  public int Calculate() => 0;
              }

              internal static class MockExtensions
              {
                  public static Mock<TInterface> AsInterface<TInterface>(Mock<SampleClass> mock)
                      where TInterface : class
                  {
                      return mock.As<TInterface>();
                  }
              }

              internal class UnitTest
              {
                  private void Test()
                  {
                      var mock = new Mock<SampleClass>();
                      var forwarded = MockExtensions.AsInterface<ISampleInterface>(mock);
                  }
              }
              """,
            referenceAssemblyGroup);
    }
```

### 4. New test — error-type negative (malformed code)

Requires `using Microsoft.CodeAnalysis.Testing;` at the top of the test file.

```csharp
    public static IEnumerable<object[]> ErrorTypeTestData()
    {
        return new object[][]
        {
            // The type argument does not resolve (mid-edit): the compiler reports CS0246;
            // Moq1300 must stay silent.
            ["""new Mock<SampleClass>().As<INonExistentInterface>();"""],
        }.WithNamespaces().WithMoqReferenceAssemblyGroups();
    }

    [Theory]
    [MemberData(nameof(ErrorTypeTestData))]
    public async Task ShouldNotReportWhenAsTypeArgumentIsUnresolved(string referenceAssemblyGroup, string @namespace, string mock)
    {
        // CompilerDiagnostics.None suppresses the intentional CS0246, mirroring the
        // malformed-code pattern used in MockRepositoryVerifyAnalyzerTests.cs:327.
        await Verifier.VerifyAnalyzerAsync(
            $$"""
              {{@namespace}}

              public class SampleClass
              {
                  public int Calculate() => 0;
              }

              internal class UnitTest
              {
                  private void Test()
                  {
                      {{mock}}
                  }
              }
              """,
            referenceAssemblyGroup,
            CompilerDiagnostics.None);
    }
```

Pre-fix, this test fails with an unexpected Moq1300 on `INonExistentInterface` — confirm red before fixing (red/green).

### 5. New test — struct/enum robustness negatives (documenting behavior)

Struct/enum type arguments violate Moq's `where TInterface : class` constraint (CS0452), and the bound operation is an `IInvalidOperation`, so the analyzer never runs on them (verified on the pinned Roslyn). Pin this with a no-crash negative using `CompilerDiagnostics.None`, expecting **no** Moq1300:

```csharp
    public static IEnumerable<object[]> ConstraintViolationTestData()
    {
        return new object[][]
        {
            // CS0452: value types cannot satisfy the `class` constraint; the invocation
            // does not bind (IInvalidOperation), so no Moq1300 is possible or expected.
            ["""new Mock<SampleClass>().As<MyStruct>();"""],
            ["""new Mock<SampleClass>().As<MyEnum>();"""],
        }.WithNamespaces().WithMoqReferenceAssemblyGroups();
    }

    [Theory]
    [MemberData(nameof(ConstraintViolationTestData))]
    public async Task ShouldNotCrashWhenAsTypeArgumentViolatesClassConstraint(string referenceAssemblyGroup, string @namespace, string mock)
    {
        await Verifier.VerifyAnalyzerAsync(
            $$"""
              {{@namespace}}

              public struct MyStruct { }

              public enum MyEnum { A }

              public class SampleClass
              {
                  public int Calculate() => 0;
              }

              internal class UnitTest
              {
                  private void Test()
                  {
                      {{mock}}
                  }
              }
              """,
            referenceAssemblyGroup,
            CompilerDiagnostics.None);
    }
```

If a future Roslyn version starts binding these as `IInvocationOperation`, this test will fail with an unexpected Moq1300 and the expectation can be consciously revisited then.

### 6. Existing tests that must keep passing

- The 4 original `TestData` rows (unchanged expectations).
- `ShouldPassIfCustomMockClassIsUsed` (doppelganger, lines 59-66) — unchanged.

## Constraints

- Analyzer assemblies target `netstandard2.0`; Roslyn pinned to Microsoft.CodeAnalysis 4.8 (ADR-003); no new package dependencies.
- Symbol-based detection (ADR-001): unchanged — the fix only adds `TypeKind` checks on an already-resolved symbol; `As` method identification stays via `MoqKnownSymbols.MockAs`/`Mock1As` (`src/Common/WellKnown/MoqKnownSymbols.cs`). The `"As"` string comparison at line 92 is a pre-existing, sanctioned diagnostic-*location* pick (not detection) and is out of scope.
- Tests follow repo patterns: `{|Moq1300:...|}` markup (ID + span), `[Theory]` + `[MemberData]`, rows multiplied across 2 namespaces and 2 Moq versions (4.8.2/4.18.4) via `WithNamespaces().WithMoqReferenceAssemblyGroups()`; malformed-code tests use `CompilerDiagnostics.None`. Moq-version sensitivity: `Mock.As<TInterface>`/`Mock<T>.As<TInterface>` exist with the same shape in both 4.8.2 and 4.18.4, so all rows run on both versions.
- Validation: `dotnet format`, `dotnet build` (0 warnings), `dotnet test --settings ./build/targets/tests/test.runsettings`.
- Update `docs/rules/Moq1300.md` (behavior change); never touch `AnalyzerReleases.Shipped.md`.

## Validation commands

```bash
dotnet format
dotnet build            # must complete with 0 warnings
dotnet test --settings ./build/targets/tests/test.runsettings
```

## Acceptance criteria

- [ ] `As<T>()` with an unresolved type argument produces no Moq1300 (test with `CompilerDiagnostics.None`; confirmed failing pre-fix).
- [ ] `As<T>()` on an open generic type parameter (`where T : class`) produces no Moq1300.
- [ ] Classes, abstract classes, and delegates (`TypeKind.Delegate`) are still flagged; the delegate row `As<{|Moq1300:System.Action|}>()` passes.
- [ ] Generic interface, nested-generic interface, chained `As`, and mock-variable rows pass as specified.
- [ ] Struct/enum constraint-violation rows produce no Moq1300 and no analyzer crash.
- [ ] The 4 pre-existing data rows and the doppelganger test pass unchanged.
- [ ] `docs/rules/Moq1300.md` documents the error-type/type-parameter exclusions; `AnalyzerReleases.Shipped.md` untouched.
- [ ] `dotnet format`, `dotnet build` (0 warnings), and the full test suite pass.

## Out of scope

- Constraint-based analysis of type parameters (flagging `As<T>() where T : SampleClass`) — documented FN trade-off.
- Changing the diagnostic-location logic (`FirstOrDefault` over descendant `GenericNameSyntax` named "As"), including its behavior on chained calls — pre-existing; pin observed spans instead.
- `Mock.Of`/other Moq1301+ rules and the audit's performance findings for other analyzers.

rjmurillo:
Fixes #1251.

## Problem
`AsShouldBeUsedOnlyForInterfaceAnalyzer` reported Moq1300 for any non-interface `TypeKind`, including open generic type parameters. That produced false positives for `As<T>()` in generic helpers where `T` may be substituted with an interface at the call site.

## Fix (constraint-aware)
Moq1300 now decides based on whether the type argument *could* be an interface:

- **Interface** type argument: never reported (valid use).
- **Unresolved (`TypeKind.Error`)**: skipped; the code already has a compiler error, so piling on is noise.
- **Open generic type parameter**: reported **only** when its constraints make an interface substitution impossible:
  - value-type constraint (`struct`/`unmanaged`) — value types are not interfaces,
  - `new()` constraint — interfaces have no constructor,
  - base-class constraint (`where T : SomeClass`) — cannot also be an interface.
  Otherwise (unconstrained, `where T : class`, or interface-constrained) it is exempt, since `T` may be an interface.

This refines the original "skip all type parameters" idea: a blanket exemption would miss cases where `T` provably cannot be an interface (raised in review by Codex).

## Tests
- `ShouldNotReportForOpenTypeParameter` (`where T : class`) — no diagnostic.
- `ShouldReportForClassConstrainedTypeParameter` (`where T : BaseSampleClass`) — reports.
- `ShouldReportForConstructorConstrainedTypeParameter` (`where T : class, new()`) — reports.

All test inputs are valid, compilable C# per `csharp.instructions.md` (no compiler-error samples).

## Docs
`docs/rules/Moq1300.md` "Advanced Scenarios: Generic Type Parameters" rewritten to describe the constraint-aware behavior with compilable examples.

## Validation
- `dotnet build`: 0 warnings, 0 errors.
- As-analyzer tests: 35 passed, 0 failed.
- Moq version: behavior identical for Moq 4.8.2 and 4.18.4 (analyzer is Moq-version-agnostic here).

### DIFF
--- a/docs/rules/Moq1300.md
+++ b/docs/rules/Moq1300.md
@@ -73,51 +73,39 @@ For more information, see
 
 ### Advanced Scenarios: Generic Type Parameters
 
-Moq1300 fires for generic type parameters constrained to interfaces, even though the constraint ensures interface compatibility at the call site:
+Whether Moq1300 fires on a generic type parameter depends on its constraints, because the
+constraints determine whether `T` could ever be an interface at a call site.
+
+Moq1300 does **not** fire when `T` could still be substituted with an interface:
 
 ```csharp
-void SetupMyMock<T>(Mock<IFoo> mock, Action<Mock<T>> setupStuff) 
+void SetupMyMock<T>(Mock<IFoo> mock, Action<Mock<T>> setupStuff)
     where T : IFoo
 {
-    // Moq1300 fires here, even though T is constrained to IFoo
+    // No Moq1300: T is constrained to the interface IFoo, so T may be an interface.
     setupStuff(mock.As<T>());
 }
 ```
 
-#### Why This Happens
+This also applies to unconstrained type parameters and those constrained with `where T : class`,
+since an interface is a reference type.
 
-C# generic constraints cannot enforce "interface-only" requirements. While `where T : IFoo` ensures `T` implements the interface `IFoo`, it doesn't prevent `T` from being a class (e.g., `class Foo : IFoo {}`), which would cause `Mock.As<T>()` to fail at runtime.
-
-Because `Mock.As<T>()` only works with interfaces, the analyzer **errs on the side of safety** and reports this as an error to prevent potential runtime failures.
-
-#### When Suppression is Appropriate
-
-- You control all call sites and can verify `T` is always an interface
-- Your helper method is internal to your test project with limited usage
-- You have integration tests validating the generic patterns
-
-#### When to Avoid Suppression
-
-- The helper is in a public library
-- Call sites are outside your control
-- Runtime failures would be difficult to diagnose
-
-##### Suppression Example
+Moq1300 **does** fire when the constraints make an interface substitution impossible, because
+`Mock.As<T>()` can then never bind to an interface:
 
 ```csharp
-void SetupMyMock<T>(Mock<IFoo> mock, Action<Mock<T>> setupStuff) 
-    where T : IFoo
+void SetupMyMock<T>(Mock<T> mock)
+    where T : SampleClass // base-class constraint: T can never be an interface
 {
-    #pragma warning disable Moq1300 // Helper method used in test where call sites are under our control
-    setupStuff(mock.As<T>());
-    #pragma warning restore Moq1300
+    mock.As<T>(); // Moq1300: Mock.As() should take interfaces only
 }
 ```
 
-> **Note:** Per [issue #756](https://github.com/rjmurillo/moq.analyzers/issues/756), this guidance may be reconsidered if:
->
->1. multiple independent users report the scenario,
->2. official Moq documentation includes this pattern,
->3. C# adds interface-only constraints,
->4. significant community validation, or
->5. the pattern appears in other major OSS projects
+A value-type constraint (`where T : struct` or `where T : unmanaged`) is likewise reported,
+though such code also fails to compile because `Mock.As<T>()` requires a reference type.
+
+> **Note:** Per [issue #756](https://github.com/rjmurillo/moq.analyzers/issues/756) and
+> [issue #1251](https://github.com/rjmurillo/moq.analyzers/issues/1251), the analyzer no longer
+> reports interface-constrained or unconstrained type parameters, which removes the previous false
+> positives for those patterns.
+
--- a/docs/rules/Moq1300.md
+++ b/docs/rules/Moq1300.md
@@ -80,7 +80,7 @@ Moq1300 does **not** fire when `T` could still be substituted with an interface:
 
 ```csharp
 void SetupMyMock<T>(Mock<IFoo> mock, Action<Mock<T>> setupStuff)
-    where T : IFoo
+    where T : class, IFoo
 {
     // No Moq1300: T is constrained to the interface IFoo, so T may be an interface.
     setupStuff(mock.As<T>());
@@ -101,8 +101,19 @@ void SetupMyMock<T>(Mock<T> mock)
 }
 ```
 
-A value-type constraint (`where T : struct` or `where T : unmanaged`) is likewise reported,
-though such code also fails to compile because `Mock.As<T>()` requires a reference type.
+A `new()` constraint is likewise reported, because an interface has no constructor and can never
+satisfy it:
+
+```csharp
+void SetupMyMock<T>(Mock<T> mock)
+    where T : class, new() // T needs a constructor, so it can never be an interface
+{
+    mock.As<T>(); // Moq1300: Mock.As() should take interfaces only
+}
+```
+
+A value-type constraint (`where T : struct` or `where T : unmanaged`) does not compile at all,
+because `Mock.As<T>()` requires a reference type.
 
 > **Note:** Per [issue #756](https://github.com/rjmurillo/moq.analyzers/issues/756) and
 > [issue #1251](https://github.com/rjmurillo/moq.analyzers/issues/1251), the analyzer no longer
--- a/src/Analyzers/AsShouldBeUsedOnlyForInterfaceAnalyzer.cs
+++ b/src/Analyzers/AsShouldBeUsedOnlyForInterfaceAnalyzer.cs
@@ -83,7 +83,14 @@ public class AsShouldBeUsedOnlyForInterfaceAnalyzer : DiagnosticAnalyzer
             return;
         }
 
-        if (typeArguments[0] is ITypeSymbol { TypeKind: not TypeKind.Interface } typeSymbol)
+        // Exclude type kinds we cannot classify as "not an interface":
+        //   - TypeParameter: an open generic (for example As<T>() in a generic method) may be
+        //     constrained to, or substituted with, an interface at the call site. Reporting here
+        //     is a false positive (issue #1251).
+        //   - Error: the type failed to bind, so the code already has a compiler error; adding
+        //     Moq1300 on top is noise.
+        if (typeArguments[0] is ITypeSymbol { TypeKind: not TypeKind.Interface } typeSymbol
+            && typeSymbol.TypeKind is not (TypeKind.TypeParameter or TypeKind.Error))
         {
             // Find the first As<T> generic type argument and report the diagnostic on it
             GenericNameSyntax? asGeneric = invocationOperation.Syntax
--- a/src/Analyzers/AsShouldBeUsedOnlyForInterfaceAnalyzer.cs
+++ b/src/Analyzers/AsShouldBeUsedOnlyForInterfaceAnalyzer.cs
@@ -83,24 +83,62 @@ public class AsShouldBeUsedOnlyForInterfaceAnalyzer : DiagnosticAnalyzer
             return;
         }
 
-        // Exclude type kinds we cannot classify as "not an interface":
-        //   - TypeParameter: an open generic (for example As<T>() in a generic method) may be
-        //     constrained to, or substituted with, an interface at the call site. Reporting here
-        //     is a false positive (issue #1251).
-        //   - Error: the type failed to bind, so the code already has a compiler error; adding
-        //     Moq1300 on top is noise.
-        if (typeArguments[0] is ITypeSymbol { TypeKind: not TypeKind.Interface } typeSymbol
-            && typeSymbol.TypeKind is not (TypeKind.TypeParameter or TypeKind.Error))
-        {
-            // Find the first As<T> generic type argument and report the diagnostic on it
-            GenericNameSyntax? asGeneric = invocationOperation.Syntax
-                .DescendantNodes()
-                .OfType<GenericNameSyntax>()
-                .FirstOrDefault(x => string.Equals(x.Identifier.ValueText, "As", StringComparison.Ordinal));
+        ITypeSymbol typeSymbol = typeArguments[0];
 
-            TypeSyntax? typeArg = asGeneric?.TypeArgumentList.Arguments.FirstOrDefault();
-            Location location = typeArg?.GetLocation() ?? invocationOperation.Syntax.GetLocation();
-            context.ReportDiagnostic(location.CreateDiagnostic(Rule, typeSymbol.ToDisplayString()));
+        // Interface: this is the valid, intended use of As<T>, so never report.
+        // Error: the type failed to bind, so the code already has a compiler error; adding
+        // Moq1300 on top is noise (issue #1251).
+        if (typeSymbol.TypeKind is TypeKind.Interface or TypeKind.Error)
+        {
+            return;
         }
+
+        // Open generic type parameter: at the call site T may be substituted with an interface,
+        // so reporting is a false positive (issue #1251) UNLESS its constraints make an interface
+        // substitution impossible (a value-type or base-class constraint). In that case As<T> can
+        // never bind to an interface, so the diagnostic is correct.
+        if (typeSymbol is ITypeParameterSymbol typeParameter && CanBeSubstitutedWithInterface(typeParameter))
+        {
+            return;
+        }
+
+        // Find the first As<T> generic type argument and report the diagnostic on it
+        GenericNameSyntax? asGeneric = invocationOperation.Syntax
+            .DescendantNodes()
+            .OfType<GenericNameSyntax>()
+            .FirstOrDefault(x => string.Equals(x.Identifier.ValueText, "As", StringComparison.Ordinal));
+
+        TypeSyntax? typeArg = asGeneric?.TypeArgumentList.Arguments.FirstOrDefault();
+        Location location = typeArg?.GetLocation() ?? invocationOperation.Syntax.GetLocation();
+        context.ReportDiagnostic(location.CreateDiagnostic(Rule, typeSymbol.ToDisplayString()));
+    }
+
+    /// <summary>
+    /// Determines whether an open generic type parameter could be substituted with an interface
+    /// at a call site, given its declared constraints.
+    /// </summary>
+    /// <param name="typeParameter">The open generic type parameter used as the <c>As&lt;T&gt;</c> argument.</param>
+    /// <returns>
+    /// <see langword="false" /> when a value-type constraint (<c>struct</c>/<c>unmanaged</c>) or a
+    /// base-class constraint forbids an interface substitution; otherwise <see langword="true" />.
+    /// </returns>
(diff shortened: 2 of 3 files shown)


rjmurillo:
## Review feedback addressed — commit 46c0097

**Class/value-type-constrained type parameters must still report (Codex P2)**
The blanket `TypeParameter` exemption over-suppressed. New `CanBeSubstitutedWithInterface` check: a value-type constraint (`struct`/`unmanaged`) or a base-class constraint means `T` can never be an interface, so Moq1300 still fires. Only unconstrained, `where T : class`, or interface-constrained type parameters are exempt (those may be substituted with an interface). Regression test `ShouldReportForClassConstrainedTypeParameter` added.

**Stale docs (Codex P2)**
`docs/rules/Moq1300.md` "Advanced Scenarios: Generic Type Parameters" rewritten to describe the constraint-aware behavior (interface/class-constrained and unconstrained → no diagnostic; base-class/value-type constrained → diagnostic).

**No test for `TypeKind.Error` path (Copilot)**
Added `ShouldNotReportForUnresolvedType`, which uses an unresolvable type argument and runs the verifier with `CompilerDiagnostics.None`.

**Pattern-match simplification (Gemini medium)**
The reporting logic was restructured with early returns and a dedicated constraint helper, which is cleaner than folding all `TypeKind` checks into a single property pattern and also accommodates the new constraint logic.

As-analyzer tests: **35 passed, 0 failed**. Build: 0 warnings.

rjmurillo:
## Review round 2 addressed — commit 606321f

**new() constraint (Codex P2)** — added: a `new()` constraint means T needs a constructor, which no interface can satisfy, so Moq1300 now fires. Test `ShouldReportForConstructorConstrainedTypeParameter` added.

**Unresolved-type test violates repo contract (Codex P1)** — correct. `csharp.instructions.md:13` forbids compiler-error test inputs. Removed `ShouldNotReportForUnresolvedType` (it used `As<DoesNotExist>()`). The `TypeKind.Error` exemption stays as defensive code; it is intentionally untestable under the valid-code-only contract.

**PR description mismatch (Copilot)** — updated the PR description to describe the constraint-aware behavior. The refined behavior (report when T provably cannot be an interface) is intentional and was requested by Codex; the class-constrained test is correct and stays.

**Docs examples don't compile (Copilot)** — fixed: `As<T>` samples now include the `class` constraint, and the value-type note no longer asserts analyzer behavior on non-compilable code.

Build: 0 warnings. As-analyzer tests: 35 passed.

PR #1295, opened 2026-07-05
fix(codefix): guard MockBehavior fixer against stale diagnostic shape (#1257)

rjmurillo:
## Summary

Fixes part (a) of #1257: `SetExplicitMockBehaviorCodeAction.GetChangedDocumentAsync` applied the recorded edit (Insert/Replace argument) without checking that the resolved node still matched the shape the analyzer saw at analysis time. A stale lightbulb (document edited between analysis and fix application) could resolve to a non-invocation/creation node, or an out-of-range position, throwing from the argument rewriters.

Part (b) of #1257 (undefined `EditType`) already shipped in #1288.

## Change

- Add `IsEditShapeValid()` guard in `SetExplicitMockBehaviorCodeAction`. It validates the node kind (invocation / object-creation) and that `_position` is within the current argument count for the requested `EditType`. Returns the original document unchanged when the shape no longer matches, instead of letting the rewriters throw.

## Tests

New direct fixer unit tests (`SetExplicitMockBehaviorFixerTests`):
- `[Theory]` stale edit-shape rows (Replace/0 empty list, Replace/5 and Insert/5 out-of-range) against the object-creation node.
- `[Fact]` stale node-kind (diagnostic span resolves to a `LocalDeclarationStatementSyntax`, not an invocation/creation).

Both assert the fixer registers its two actions and applying either leaves the document unchanged (graceful no-op). The standard analyzer/code-fix harness cannot reach this path because the analyzer always emits edit properties matching the node it analyzed, so the tests synthesize the stale diagnostic directly.

Positive/negative/boundary coverage of `IsEditShapeValid`: node-kind mismatch, Insert range boundary, Replace range boundary. The `EditType` `_ => false` default is compiler-mandated switch-exhaustiveness and unreachable (EditType validated upstream in #1288); accepted exclusion.

## Validation

- `dotnet build ... /p:ContinuousIntegrationBuild=true` (PedanticMode): 0 warnings, 0 errors.
- `dotnet test --filter SetExplicitMockBehaviorFixer`: 4 passed.
- Behavior-only fix: no `docs/rules/` or `AnalyzerReleases` changes required.

## Moq version

No Moq API surface change; targets the same Moq versions as the existing fixer.

Related to #1257

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved code action safety by validating that the diagnostic edit still matches the current code shape before applying changes.
  * Prevented fixes from running on unsupported/stale patterns, avoiding unintended document modifications.
  * For insert scenarios, now avoids adding a duplicate `MockBehavior` argument when it’s already present.

* **Tests**
  * Added/extended coverage for stale or mismatched diagnostics to ensure code actions no-op safely.
  * Confirmed applying the code actions leaves the document unchanged (no text changes, no exceptions), including duplicate-prevention cases.

### DIFF
--- a/src/CodeFixes/SetExplicitMockBehaviorCodeAction.cs
+++ b/src/CodeFixes/SetExplicitMockBehaviorCodeAction.cs
@@ -59,6 +59,15 @@ internal sealed class SetExplicitMockBehaviorCodeAction : CodeAction
             return _document;
         }
 
+        // Defense in depth: the edit properties attached to the diagnostic describe the argument-list
+        // shape the analyzer saw at analysis time. If the document changed between analysis and fix
+        // application (stale lightbulb), the resolved node may no longer match that shape. Return the
+        // original document instead of letting the argument rewriters throw.
+        if (!IsEditShapeValid())
+        {
+            return _document;
+        }
+
         SyntaxNode behavior = _behaviorType switch
         {
             BehaviorType.Loose => editor.Generator.MemberAccessExpression(knownSymbols.MockBehaviorLoose),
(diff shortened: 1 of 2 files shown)


rjmurillo:
Adversarial review (GPT-5.5) follow-up: the Insert-duplicate risk it flagged is now fixed in commit 8a05306 (`CanApplyEdit` verifies no current argument resolves to `MockBehavior` before inserting, with a regression test). The narrower Replace same-type drift is tracked separately in #1296 to avoid false-negatives on valid multi-part expressions.

rjmurillo:
Codacy Static (non-required check) flags `AssertFixerNoOpAsync` at SetExplicitMockBehaviorFixerTests.cs:148 with "Remove the 'Async' suffix." Keeping the suffix: the method is a genuine `async Task` helper that awaits `RegisterCodeFixesAsync`/`GetOperationsAsync`/`GetTextAsync`, so the suffix is correct per the .NET TAP naming guidelines and matches the dominant convention in this test project (`VerifyCodeFixAsync`, `VerifyBothAnalyzersAsync`, `VerifyAnalyzerAsync`, `VerifyAllAnalyzersAsync`). Renaming would violate that convention. The `AssemblyVersion` annotation at :1 is pre-existing infra noise present on every test file. The actionable production finding from the prior run (CanApplyEdit cyclomatic complexity 13) was fixed in 2ca9263 by extracting `TryGetArgumentList` and `ContainsMockBehaviorArgument`; Codacy Static now reports 0 production issues.

PR #1302, opened 2026-07-06
fix(analyzers): Moq1210 handle named-argument reordering in Verify

### LINKED ISSUE #1263: Moq1210 false negative: named-argument reordering in Verify/VerifyGet/VerifySet skips analysis because helpers index Arguments[0] positionally
## Summary

`VerifyShouldBeUsedOnlyForOverridableMembersAnalyzer` (rule **Moq1210**) and the shared `MoqVerificationHelpers` extract the verification lambda by indexing `IInvocationOperation.Arguments[0]`. `Arguments` is in **evaluation (source) order**, not parameter order, so a call using C# named-argument reordering —

```csharp
mock.Verify(times: Times.Once(), expression: x => x.NonVirtualMethod());
```

— puts the `times` argument at index 0. `ExtractLambdaFromArgument` then fails to find a lambda, the helper returns `null`, and the analyzer silently skips the invocation: a **false negative** on a call that throws `NotSupportedException` at runtime ("Non-overridable members ... may not be used in setup / verification expressions").

The repository already solves this exact problem correctly in `NoMethodsInPropertySetupAnalyzer` (`src/Analyzers/NoMethodsInPropertySetupAnalyzer.cs:82-92`), which selects the argument whose `Parameter?.Ordinal == 0`. Replicate that pattern.

**Verified against both Moq versions the test suite pins** (reflected over `Moq.dll` 4.8.2 and 4.18.4): the named form is legal in both. Overload parameter names are identical across the two versions:

- `Mock<T>.Verify(expression)` / `Verify(expression, times)` / `Verify(expression, Func<Times> times)` / `Verify(expression, failMessage)` / `Verify(expression, times, failMessage)`
- `Mock<T>.VerifyGet(expression, ...)` — same names.
- `Mock<T>.VerifySet(setterExpression, ...)` — note the first parameter is named **`setterExpression`**, not `expression` (same in 4.8.2 and 4.18.4); use that name in VerifySet test rows.

### `src/Analyzers/VerifyShouldBeUsedOnlyForOverridableMembersAnalyzer.cs:64-75`

```csharp
        ISymbol? mockedMemberSymbol;
        if (targetMethod.IsInstanceOf(knownSymbols.Mock1VerifySet))
        {
            IAnonymousFunctionOperation? lambda = MoqVerificationHelpers.ExtractLambdaFromArgument(invocationOperation.Arguments[0].Value);
            mockedMemberSymbol = lambda is not null
                ? MoqVerificationHelpers.ExtractPropertyFromVerifySetLambda(lambda)
                : null;
        }
        else
        {
            mockedMemberSymbol = MoqVerificationHelpers.TryGetMockedMemberSymbol(invocationOperation);
        }
```

### `src/Common/MoqVerificationHelpers.cs:58-89` (both `Arguments[0]` sites)

```csharp
    public static ISymbol? TryGetMockedMemberSymbol(IInvocationOperation moqInvocation)
    {
        if (moqInvocation.Arguments.Length == 0)
        {
            return null;
        }

        // For code fix, we only need to handle the simple case (first argument)
        // since the analyzer already validates the more complex scenarios
        IAnonymousFunctionOperation? lambdaOperation = ExtractLambdaFromArgument(moqInvocation.Arguments[0].Value);

        return lambdaOperation?.Body.GetReferencedMemberSymbolFromLambda();
    }
```

and the identical shape in `TryGetMockedMemberSyntax` at `:77-89` (indexing at `:86`).

### The correct in-repo pattern (`src/Analyzers/NoMethodsInPropertySetupAnalyzer.cs:82-92`)

```csharp
        // Extract the lambda argument by parameter ordinal (0), not source position,
        // so named-argument reordering (e.g., SetupProperty(initialValue: ..., propertyExpression: ...)) is handled.
        IArgumentOperation? lambdaArgument = null;
        foreach (IArgumentOperation argument in invocationOperation.Arguments)
        {
            if (argument.Parameter?.Ordinal == 0)
            {
                lambdaArgument = argument;
                break;
            }
        }
```

### Other `Arguments[0]`-by-position sites checked (completeness)

- `src/Analyzers/MethodSetupShouldSpecifyReturnValueAnalyzer.cs:134` (`moqSetupInvocation.Arguments[0].Value`) — the matched `Setup` methods take a single parameter, and even `Setup(expression: ...)` keeps the lone argument at index 0, so it is not analysis-affecting; leave it (optionally note it for a follow-up consistency pass).
- `src/Common/SemanticModelExtensions.cs:60` and `src/Common/InvocationExpressionSyntaxExtensions.cs:62` index **syntax** argument lists for `Setup(...)` (single-parameter, same reasoning) and are additionally covered by audit finding A-2 (crash guard) — out of scope here.
- There are no other `IInvocationOperation.Arguments[0]` sites in `src/` (grep-verified).

Fixing the two `MoqVerificationHelpers` members also hardens the other callers of `TryGetMockedMemberSymbol`/`TryGetMockedMemberSyntax` (`SetupShouldBeUsedOnlyForOverridableMembersAnalyzer.cs:90`, `SetupSequenceShouldBeUsedOnlyForOverridableMembersAnalyzer.cs:62`, `:71`, `src/CodeFixes/VerifyOverridableMembersFixer.cs:57`) — for `Verify*` shapes reached via the fixer this is load-bearing; for single-parameter `Setup`/`SetupSequence` it is a no-op.

## Desired behavior

`mock.Verify(times: Times.Once(), expression: x => x.NonVirtualMethod())` (and the `VerifyGet`/`VerifySet` named forms) produce the same Moq1210 diagnostic as the positional form. Positional calls are untouched. Compiler-synthesized default-value arguments still resolve correctly because `IArgumentOperation.Parameter` is populated for them too.

### 1. `src/Common/MoqVerificationHelpers.cs` — add an ordinal-based accessor and use it in both members

Add:

```csharp
    /// <summary>
    /// Gets the argument bound to the parameter at the given ordinal, regardless of source order.
    /// </summary>
    /// <remarks>
    /// <see cref="IInvocationOperation.Arguments"/> is in evaluation (source) order, so named-argument
    /// reordering (e.g., <c>Verify(times: ..., expression: ...)</c>) moves the expression argument away
    /// from index 0. Selecting by <see cref="IArgumentOperation.Parameter"/> ordinal mirrors
    /// NoMethodsInPropertySetupAnalyzer (src/Analyzers/NoMethodsInPropertySetupAnalyzer.cs:82-92).
    /// </remarks>
    /// <param name="moqInvocation">The invocation operation.</param>
    /// <param name="ordinal">The zero-based parameter ordinal.</param>
    /// <returns>The matching argument, or <see langword="null"/> if none is bound to that ordinal.</returns>
    public static IArgumentOperation? GetArgumentForParameterOrdinal(IInvocationOperation moqInvocation, int ordinal)
    {
        foreach (IArgumentOperation argument in moqInvocation.Arguments)
        {
            if (argument.Parameter?.Ordinal == ordinal)
            {
                return argument;
            }
        }

        return null;
    }
```

Replace both extractors:

```csharp
    /// <summary>
    /// Extracts the mocked member symbol from a Moq Verify/Setup invocation.
    /// </summary>
    /// <param name="moqInvocation">The invocation operation.</param>
    /// <returns>The mocked member symbol if found, otherwise null.</returns>
    public static ISymbol? TryGetMockedMemberSymbol(IInvocationOperation moqInvocation)
    {
        // Select the expression argument by parameter ordinal (0), not source position,
        // so named-argument reordering (e.g., Verify(times: ..., expression: ...)) is handled.
        IArgumentOperation? expressionArgument = GetArgumentForParameterOrdinal(moqInvocation, 0);
        if (expressionArgument is null)
        {
            return null;
        }

        IAnonymousFunctionOperation? lambdaOperation = ExtractLambdaFromArgument(expressionArgument.Value);

        return lambdaOperation?.Body.GetReferencedMemberSymbolFromLambda();
    }

    /// <summary>
    /// Extracts the mocked member syntax node from a Moq Verify/Setup invocation.
    /// </summary>
    /// <param name="moqInvocation">The invocation operation.</param>
    /// <returns>The mocked member syntax node if found, otherwise null.</returns>
    public static SyntaxNode? TryGetMockedMemberSyntax(IInvocationOperation moqInvocation)
    {
        // Select the expression argument by parameter ordinal (0), not source position,
        // so named-argument reordering (e.g., Verify(times: ..., expression: ...)) is handled.
        IArgumentOperation? expressionArgument = GetArgumentForParameterOrdinal(moqInvocation, 0);
        if (expressionArgument is null)
        {
            return null;
        }

        IAnonymousFunctionOperation? lambdaOperation = ExtractLambdaFromArgument(expressionArgument.Value);

        return lambdaOperation?.Body.GetReferencedMemberSyntaxFromLambda();
    }
```

(The `Arguments.Length == 0` pre-checks become redundant — the loop returns `null` for empty arrays — and are removed.)

### 2. `src/Analyzers/VerifyShouldBeUsedOnlyForOverridableMembersAnalyzer.cs` — fix the VerifySet branch

Replace `AnalyzeInvocation` (`:50-88`):

```csharp
    private static void AnalyzeInvocation(OperationAnalysisContext context, MoqKnownSymbols knownSymbols)
    {
        if (context.Operation is not IInvocationOperation invocationOperation)
        {
            return;
        }

        IMethodSymbol targetMethod = invocationOperation.TargetMethod;

        if (!ShouldAnalyzeMethod(targetMethod, knownSymbols))
        {
            return;
        }

        ISymbol? mockedMemberSymbol;
        if (targetMethod.IsInstanceOf(knownSymbols.Mock1VerifySet))
        {
            // Select the setter expression by parameter ordinal (0), not source position,
            // so named-argument reordering (e.g., VerifySet(times: ..., setterExpression: ...)) is handled.
            IArgumentOperation? setterArgument = MoqVerificationHelpers.GetArgumentForParameterOrdinal(invocationOperation, 0);
            IAnonymousFunctionOperation? lambda = setterArgument is not null
                ? MoqVerificationHelpers.ExtractLambdaFromArgument(setterArgument.Value)
                : null;
            mockedMemberSymbol = lambda is not null
                ? MoqVerificationHelpers.ExtractPropertyFromVerifySetLambda(lambda)
                : null;
        }
        else
        {
            mockedMemberSymbol = MoqVerificationHelpers.TryGetMockedMemberSymbol(invocationOperation);
        }

        if (mockedMemberSymbol == null || IsInterfaceMember(mockedMemberSymbol))
        {
            return;
        }

        if (IsMemberAllowedForVerification(mockedMemberSymbol, targetMethod, knownSymbols))
        {
            return;
        }

        ReportDiagnostic(context, invocationOperation, mockedMemberSymbol);
    }
```

All other members of the analyzer are unchanged.

## Tests

Extend `tests/Moq.Analyzers.Test/VerifyShouldBeUsedOnlyForOverridableMembersAnalyzerTests.cs`, mirroring its exact conventions (rows in `TestData()` with `{|Moq1210:...|}` markup, `.WithNamespaces().WithMoqReferenceAssemblyGroups()` for both Moq versions, `.WithNewMoqReferenceAssemblyGroups()` for VerifySet — matching the existing split at `:10-41`; the diagnostic spans the whole invocation because `ReportDiagnostic` uses `invocationOperation.Syntax`):

Add to the `both` rows (parameter names `expression`/`times` verified identical in Moq 4.8.2 and 4.18.4):

```csharp
            // Named-argument reordering must not skip analysis (regression: Arguments[0] was positional)
            ["""{|Moq1210:new Mock<SampleClass>().Verify(times: Times.Once(), expression: x => x.Calculate(It.IsAny<int>(), It.IsAny<int>(), It.IsAny<int>()))|};"""],
            ["""{|Moq1210:new Mock<SampleClass>().VerifyGet(times: Times.Once(), expression: x => x.Property)|};"""],

            // Named-argument reordering on an overridable member stays clean (negative)
            ["""new Mock<BaseSampleClass>().Verify(times: Times.Once(), expression: x => x.Calculate(It.IsAny<int>(), It.IsAny<int>()));"""],
```

Add to the `newMoqOnly` rows (VerifySet's first parameter is named `setterExpression` — verified in both Moq versions, but VerifySet rows live in the new-Moq group per the existing file's convention at `:33-39`):

```csharp
            // Named-argument reordering for VerifySet (parameter is `setterExpression`, not `expression`)
            ["""{|Moq1210:new Mock<SampleClass>().VerifySet(times: Times.Once(), setterExpression: x => { x.Property = It.IsAny<int>(); })|};"""],
```

Before implementing the fix, confirm the two positive rows FAIL on the current analyzer (missing expected diagnostic) — that proves they reproduce the false negative. All existing rows must pass unchanged (positional calls hit the ordinal-0 argument at index 0, so behavior is identical).

## Constraints

- Analyzer assemblies target **netstandard2.0**; Roslyn pinned to **Microsoft.CodeAnalysis 4.8** (ADR-003). `IArgumentOperation.Parameter`/`Ordinal` are long-standing APIs. **No new package dependencies.**
- **ADR-001** (name-based checks allowed only as fast-path pre-filters before an authoritative symbol check): this change adds **no** name-based logic — argument selection is by bound parameter symbol ordinal, which is the symbol-based approach ADR-001 wants; stated because sibling issues in this audit batch do add (sanctioned) name pre-filters.
- **ADR-007**: analyzer already uses `RegisterOperationAction` (`:45-47`); unchanged.
- Tests are mandatory and specified above.

## Validation

1. `dotnet format` (no diff).
2. `dotnet build` — **0 warnings**.
3. `dotnet test --settings ./build/targets/tests/test.runsettings` — all tests pass, including the new named-argument rows against both Moq 4.8.2 and 4.18.4 reference groups.
4. Perf: `tests/Moq.Analyzers.Benchmarks/Moq1210VerifyOverridableMembersBenchmarks.cs` exists; run the documented flow (`build/scripts/perf/README.md`; CONTRIBUTING.md "Performance Testing Guidelines", line 811): `./build/scripts/perf/PerfCore.ps1 -projects "tests/Moq.Analyzers.Benchmarks/Moq.Analyzers.Benchmarks.csproj" -diff`. Expect neutral (a short loop replaces an index).

## Acceptance criteria

- [ ] `mock.Verify(times: ..., expression: x => x.NonOverridable...)` and the `VerifyGet`/`VerifySet` named forms report Moq1210 with the same span/message as positional calls.
- [ ] Named-argument calls on overridable/interface members remain diagnostic-free.
- [ ] Both `MoqVerificationHelpers` extractors and the analyzer's VerifySet branch select arguments by `Parameter?.Ordinal == 0`; no `IInvocationOperation.Arguments[0]` positional indexing remains in these three sites.
- [ ] All existing tests pass unchanged; build 0 warnings.

## Out of scope

- Refactoring `NoMethodsInPropertySetupAnalyzer.cs:82-92` to use the new shared helper (behavior-identical cleanup; fine as a follow-up, not required here).
- `MethodSetupShouldSpecifyReturnValueAnalyzer.cs:134` and the syntax-level `Arguments[0]` sites in `SemanticModelExtensions.cs:60`/`InvocationExpressionSyntaxExtensions.cs:62` (single-parameter `Setup`; the crash aspect of `SemanticModelExtensions.cs:60` is audit finding A-2 with its own issue).
- Moq1200/Moq1207 named-argument test rows (they benefit automatically via the shared helper, but their `Setup`/`SetupSequence` methods are single-parameter, so no user-visible change exists to test).

rjmurillo:
## Summary

Fixes #1263. `MoqVerificationHelpers` extracted the expression lambda by positional `Arguments[0]`. Roslyn's `IInvocationOperation.Arguments` is in **source order**, not parameter order, so a reordered call like `mock.Verify(times: Times.Once(), expression: x => x.DoSomething())` put `Times.Once()` at index 0. `ExtractLambdaFromArgument` then returned null and Moq1210 (VerifyShouldBeUsedOnlyForOverridableMembers) silently missed the setter/getter/method — a false negative. Empirically verified with a Roslyn probe.

## Changes

- Add `MoqVerificationHelpers.GetArgumentForParameterOrdinal` that matches by `IArgumentOperation.Parameter.Ordinal`, independent of source order.
- `TryGetMockedMemberSymbol`, `TryGetMockedMemberSyntax`, and the `VerifySet` path in `VerifyShouldBeUsedOnlyForOverridableMembersAnalyzer` now use it.
- `Debug.Assert` guards the non-negative-ordinal and bound-parameter invariants.

## Tests

Positive (non-overridable member flagged), negative (overridable/interface not flagged), boundary (single-lambda, multi-named reorder, missing ordinal), for `Verify`/`VerifyGet`/`VerifySet`, plus direct unit tests for the new helper.

## Validation

- `dotnet build /p:PedanticMode=true`: 0 warnings, 0 errors.
- Full pre-push build + test suite passed. Verify filter: 251 passed; Setup filter: 770 passed.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of named arguments in verification calls, so mocked-member detection and overridability checks work correctly even when arguments are supplied in a different order.
  * Corrected `VerifySet` behavior for verification expressions provided via named-parameter overloads.
* **Tests**
  * Added unit tests covering named-argument reordering scenarios and missing/absent ordinal arguments.
  * Expanded analyzer test coverage for additional `Verify`/`VerifyGet`/`VerifySet` overload patterns (including `times:` and related variants).

### DIFF
--- a/src/Analyzers/VerifyShouldBeUsedOnlyForOverridableMembersAnalyzer.cs
+++ b/src/Analyzers/VerifyShouldBeUsedOnlyForOverridableMembersAnalyzer.cs
@@ -64,7 +64,10 @@ public class VerifyShouldBeUsedOnlyForOverridableMembersAnalyzer : DiagnosticAna
         ISymbol? mockedMemberSymbol;
         if (targetMethod.IsInstanceOf(knownSymbols.Mock1VerifySet))
         {
-            IAnonymousFunctionOperation? lambda = MoqVerificationHelpers.ExtractLambdaFromArgument(invocationOperation.Arguments[0].Value);
+            IArgumentOperation? setterArgument = MoqVerificationHelpers.GetArgumentForParameterOrdinal(invocationOperation, 0);
+            IAnonymousFunctionOperation? lambda = setterArgument is not null
+                ? MoqVerificationHelpers.ExtractLambdaFromArgument(setterArgument.Value)
+                : null;
             mockedMemberSymbol = lambda is not null
                 ? MoqVerificationHelpers.ExtractPropertyFromVerifySetLambda(lambda)
                 : null;
--- a/src/Common/MoqVerificationHelpers.cs
+++ b/src/Common/MoqVerificationHelpers.cs
@@ -57,14 +57,11 @@ internal static class MoqVerificationHelpers
     /// <returns>The mocked member symbol if found, otherwise null.</returns>
     public static ISymbol? TryGetMockedMemberSymbol(IInvocationOperation moqInvocation)
     {
-        if (moqInvocation.Arguments.Length == 0)
-        {
-            return null;
-        }
+        IArgumentOperation? expressionArgument = GetArgumentForParameterOrdinal(moqInvocation, 0);
 
-        // For code fix, we only need to handle the simple case (first argument)
-        // since the analyzer already validates the more complex scenarios
-        IAnonymousFunctionOperation? lambdaOperation = ExtractLambdaFromArgument(moqInvocation.Arguments[0].Value);
+        IAnonymousFunctionOperation? lambdaOperation = expressionArgument is not null
+            ? ExtractLambdaFromArgument(expressionArgument.Value)
+            : null;
 
         return lambdaOperation?.Body.GetReferencedMemberSymbolFromLambda();
     }
@@ -76,15 +73,35 @@ internal static class MoqVerificationHelpers
     /// <returns>The mocked member syntax node if found, otherwise null.</returns>
     public static SyntaxNode? TryGetMockedMemberSyntax(IInvocationOperation moqInvocation)
     {
-        if (moqInvocation.Arguments.Length == 0)
-        {
-            return null;
-        }
+        IArgumentOperation? expressionArgument = GetArgumentForParameterOrdinal(moqInvocation, 0);
 
-        // For code fix, we only need to handle the simple case (first argument)
-        // since the analyzer already validates the more complex scenarios
-        IAnonymousFunctionOperation? lambdaOperation = ExtractLambdaFromArgument(moqInvocation.Arguments[0].Value);
+        IAnonymousFunctionOperation? lambdaOperation = expressionArgument is not null
+            ? ExtractLambdaFromArgument(expressionArgument.Value)
+            : null;
 
         return lambdaOperation?.Body.GetReferencedMemberSyntaxFromLambda();
+    }
+
+    /// <summary>
+    /// Gets the argument bound to the parameter at the given ordinal, regardless of source order.
+    /// </summary>
+    /// <param name="moqInvocation">The invocation operation.</param>
+    /// <param name="ordinal">The zero-based parameter ordinal.</param>
+    /// <returns>The matching argument, or null if none is bound to that ordinal.</returns>
+    public static IArgumentOperation? GetArgumentForParameterOrdinal(IInvocationOperation moqInvocation, int ordinal)
+    {
+        System.Diagnostics.Debug.Assert(ordinal >= 0, "Parameter ordinal must be non-negative.");
+
+        IArgumentOperation? match = null;
+        foreach (IArgumentOperation argument in moqInvocation.Arguments)
+        {
+            if (argument.Parameter?.Ordinal == ordinal)
+            {
+                match = argument;
+                break;
+            }
+        }
+
+        return match;
     }
 }
--- a/tests/Moq.Analyzers.Test/Common/MoqVerificationHelpersTests.cs
+++ b/tests/Moq.Analyzers.Test/Common/MoqVerificationHelpersTests.cs
@@ -248,6 +248,89 @@ public class Test
     }
 
     [Fact]
+    public async Task TryGetMockedMemberSymbol_NamedArgumentReordering_ReturnsMethodSymbol()
+    {
+        const string code = @"
+using Moq;
+
+public interface IFoo
+{
+    int DoSomething();
+}
+
+public class Test
+{
+    public void M()
+    {
+        var mock = new Mock<IFoo>();
+        mock.Verify(times: Times.Once(), expression: x => x.DoSomething());
+    }
+}";
+        IInvocationOperation invocation = await GetMoqInvocationAsync(code, "Verify");
+
+        ISymbol? result = MoqVerificationHelpers.TryGetMockedMemberSymbol(invocation);
+
+        Assert.NotNull(result);
+        Assert.IsAssignableFrom<IMethodSymbol>(result);
+        Assert.Equal("DoSomething", result!.Name);
+    }
+
+    [Fact]
+    public async Task GetArgumentForParameterOrdinal_NamedArgumentReordering_ReturnsExpressionArgument()
+    {
+        const string code = @"
+using Moq;
+
+public interface IFoo
+{
+    int DoSomething();
+}
+
+public class Test
+{
+    public void M()
+    {
+        var mock = new Mock<IFoo>();
+        mock.Verify(times: Times.Once(), expression: x => x.DoSomething());
+    }
+}";
+        IInvocationOperation invocation = await GetMoqInvocationAsync(code, "Verify");
+
+        IArgumentOperation? result = MoqVerificationHelpers.GetArgumentForParameterOrdinal(invocation, 0);
+
+        Assert.NotNull(result);
+        Assert.Equal(0, result!.Parameter?.Ordinal);
+        Assert.Equal("expression", result.Parameter?.Name);
+        Assert.Contains("DoSomething", result.Value.Syntax.ToString(), StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public async Task GetArgumentForParameterOrdinal_MissingOrdinal_ReturnsNull()
+    {
+        const string code = @"
+using Moq;
+
+public interface IFoo
+{
+    int DoSomething();
+}
+
+public class Test
+{
+    public void M()
+    {
+        var mock = new Mock<IFoo>();
+        mock.Verify(expression: x => x.DoSomething());
+    }
+}";
+        IInvocationOperation invocation = await GetMoqInvocationAsync(code, "Verify");
+
+        IArgumentOperation? result = MoqVerificationHelpers.GetArgumentForParameterOrdinal(invocation, 99);
+
+        Assert.Null(result);
+    }
+
+    [Fact]
     public async Task TryGetMockedMemberSyntax_ZeroArguments_ReturnsNull()
     {
         const string code = @"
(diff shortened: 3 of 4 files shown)


rjmurillo:
@coderabbitai Addressed:

1. **`<see langword="null"/>` in `<returns>`** — Declined for local consistency. Every existing `<returns>` in `MoqVerificationHelpers.cs` (lines 14, 33, 57, 73) uses plain "null". Matching the file's established convention; a single langword tag here would make the new method inconsistent with its four siblings. Can be swept file-wide in a dedicated docs pass if desired.

2. **Redundant `Debug.Assert(argument.Parameter is not null)`** — Stale. That assert was reviewed against a pre-force-push revision. The current `GetArgumentForParameterOrdinal` (commit b2d511e) has only `Debug.Assert(ordinal >= 0)`, a deliberate boundary guard on the ordinal input, not on `argument.Parameter`. No dead assert remains.

PR #1342, opened 2026-07-09
test(codefixes): cover SyntaxGeneratorExtensions guards and defensive fixer branches

### LINKED ISSUE #1341: test(codefixes): Moq.CodeFixes internals are not unit-testable; argument-validation guards and edge branches stay uncovered
## Context

A coverage/quality evaluation of the shipping DLLs (91% line / 80% branch) found the remaining sub-100% lines fall into structural buckets, not ordinary test debt. The clearest actionable cluster is in `Moq.CodeFixes`. This issue captures it so the gaps are either closed or consciously accepted.

Related: #1332 (PerfDiff.dll coverage exclusion) is a different assembly but the same "coverage can't see this code" theme.

## Root cause

`tests/Moq.Analyzers.Test/Moq.Analyzers.Test.csproj` references **only** `src/Analyzers/Moq.Analyzers.csproj` (as an analyzer package via `AddPackageAsOutput`) and imports `Common.projitems`. It has **no `ProjectReference` to `Moq.CodeFixes` and no `InternalsVisibleTo`** into it. Code-fix internals are therefore exercised **only** through the `CodeFixVerifier`, which always supplies valid inputs. Any internal helper's defensive/argument-validation branch is unreachable.

### 1. `src/CodeFixes/SyntaxGeneratorExtensions.cs` (80.0%) - argument-validation guards, unreachable via verifier
Uncovered lines `16, 35, 42, 61, 68, 76, 81` are all `throw` statements:
- `InsertArguments`: non-`ArgumentSyntax` items (16); wrong `syntax` type (35); index out of range (68).
- `ReplaceArgument`: non-`ArgumentSyntax` item (42); wrong `syntax` type (61); index out of range / empty (76, 81).

These are contract guards on an `internal` helper. The verifier never passes invalid arguments, so they cannot fire through the code-fix path.

### 2. `src/CodeFixes/MockBehaviorFixerBase.cs` (77.8%)
Uncovered lines `21, 26`: the early `return`s when `TryGetEditProperties` fails or `nodeToFix` is null - malformed-properties paths the analyzer never produces.

### 3. `src/CodeFixes/VerifyOverridableMembersFixer.cs` (83.3%) - one genuinely reachable branch
Uncovered lines `29, 43, 49, 55, 61, 106, 112, 123` are defensive guards, **but line 38** (`invocationNode = invocationNode.FirstAncestorOrSelf<InvocationExpressionSyntax>()`) is a real, reachable scenario: it runs when `FindNode` on the diagnostic span returns a node that is not directly an `InvocationExpressionSyntax`. No current test exercises the nested-node path.

## Proposed work (maintainer to decide 1 vs 2)

- [ ] **Reachable now (no architecture change):** add a `VerifyOverridableMembersFixer` code-fix test where the flagged member access is nested so the diagnostic span lands on a non-invocation inner node, covering line 38 (positive + boundary).
- [ ] **Needs decision (architecture):** add `[assembly: InternalsVisibleTo("Moq.Analyzers.Test")]` to `Moq.CodeFixes`, then add `SyntaxGeneratorExtensionsTests` covering each guard (positive/negative/boundary per repo policy). This proves the defensive contracts and closes the guard gaps.
- [ ] **Alternative for pure guards:** if IVT is unwanted, mark the intentionally-unreachable defensive branches (e.g. `[ExcludeFromCodeCoverage]` with justification, or documented) so per-class coverage reflects reality instead of implying test debt.

## Non-goals

- The unresolved-symbol name-fallback branches in analyzers (e.g. `MethodSetupShouldSpecifyReturnValueAnalyzer` lines 147, 172-182, 197) are **out of scope**: they only fire on non-compiling code, which `csharp.instructions.md` forbids in tests. They are a deliberate structural floor, not test debt.

rjmurillo:
## Summary

Closes #1341. Adds unit tests for previously-uncovered guard and defensive branches in the `Moq.CodeFixes` assembly. No production code changes; tests only.

`SyntaxGeneratorExtensions` is internal to the shipping `Moq.CodeFixes` assembly. Tests reach it via reflection on the real assembly (the same strategy as `VerifyShouldBeUsedOnlyForOverridableMembersAnalyzerTests.InvokeCanMakeMemberVirtual`), not `InternalsVisibleTo`. IVT would surface `Moq.CodeFixes`'s Polyfill-generated BCL attribute types to the net8.0 test project and cause 122 CS0433 collisions with `System.Runtime`.

### Changes

- **`SyntaxGeneratorExtensionsTests.cs`** (new): 16 positive/negative/boundary tests covering every guard in `InsertArguments` and `ReplaceArgument` (item-type, syntax-kind, index range) plus the null-`ArgumentList` object-creation branches. `SyntaxGeneratorExtensions` reaches 100% branch coverage.
- **`VerifyOverridableMembersFixerTests.cs`** (new): drives the ancestor-walk branch (`VerifyOverridableMembersFixer.cs:38`) via a synthetic diagnostic on the nested `Verify` identifier. That branch is unreachable through the real pipeline because the analyzer always reports on the whole invocation.
- **`SetExplicitMockBehaviorFixerTests.cs`** (modified): adds a case for the `MockBehaviorFixerBase` early return when a diagnostic carries no edit properties.

### Intentionally left (unreachable-defensive)

- `MockBehaviorFixerBase.cs:26` (null root) requires a document without a syntax tree, not constructible via the fixer harness.
- Remaining null-guards in `VerifyOverridableMembersFixer` are defensive against inputs the pipeline never produces.

## Validation

| Step | Result |
| ---- | ------ |
| `dotnet build -c Release` | Build succeeded, 0 warnings |
| `dotnet test` (new/modified tests) | 28 passed, 0 failed |
| Coverage (`SyntaxGeneratorExtensions`) | branch-rate 1.00; only uncovered line is the unrelated `MemberAccessExpression`, covered by other fixer tests in the full suite |
| `dotnet format --verify-no-changes` | clean |

### Moq version compatibility

No Moq API surface touched. Tests target the same Moq version as the existing test project.

## Summary by CodeRabbit

* **Tests**
  * Expanded automated coverage for code-fix behavior, including edge cases where diagnostic details are incomplete.
  * Added validation for argument editing scenarios to improve confidence in refactoring actions.
  * Added coverage for member verification fixes to ensure the expected fix is offered in the right situations.

### DIFF
--- a/tests/Moq.Analyzers.Test/SetExplicitMockBehaviorFixerTests.cs
+++ b/tests/Moq.Analyzers.Test/SetExplicitMockBehaviorFixerTests.cs
@@ -255,6 +255,47 @@ public class SetExplicitMockBehaviorFixerTests
         await AssertFirstFixAsync(model, tree, creation.Span, source, expected, "Replace", "0");
     }
 
+    [Fact]
+    public async Task MissingEditProperties_DoesNotThrow_AndRegistersNoActions()
+    {
+        (SemanticModel model, SyntaxTree tree) = await CompilationHelper.CreateMoqCompilationAsync(Source);
+        SyntaxNode root = await tree.GetRootAsync();
+
+        ObjectCreationExpressionSyntax creation = root
+            .DescendantNodes()
+            .OfType<ObjectCreationExpressionSyntax>()
+            .Single();
+
+        // A diagnostic without the recorded edit properties must be ignored: the fixer cannot know
+        // what edit to apply, so it must no-op rather than throw or register an action.
+        DiagnosticDescriptor descriptor = new(
+            DiagnosticIds.SetExplicitMockBehavior,
+            "Test",
+            "Test message",
+            "Test",
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true);
+
+        Diagnostic diagnostic = Diagnostic.Create(descriptor, Location.Create(tree, creation.Span));
+
+        using AdhocWorkspace workspace = new();
+        Project project = workspace.AddProject("TestProject", LanguageNames.CSharp)
+            .AddMetadataReferences(model.Compilation.References);
+        Document document = project.AddDocument("Test.cs", SourceText.From(Source));
+
+        List<CodeAction> actions = [];
+        CodeFixContext context = new(
+            document,
+            diagnostic,
+            (action, _) => actions.Add(action),
+            CancellationToken.None);
+
+        Moq.CodeFixes.SetExplicitMockBehaviorFixer fixer = new();
+        await fixer.RegisterCodeFixesAsync(context);
+
+        Assert.Empty(actions);
+    }
+
     private static async Task AssertFixerNoOpAsync(SemanticModel model, SyntaxTree tree, TextSpan span, string sourceText, string editType, string editPosition)
     {
         await AssertFixerAsync(
--- a/tests/Moq.Analyzers.Test/SyntaxGeneratorExtensionsTests.cs
+++ b/tests/Moq.Analyzers.Test/SyntaxGeneratorExtensionsTests.cs
@@ -1,0 +1,262 @@
+using System.Reflection;
+using Microsoft.CodeAnalysis.Editing;
+
+namespace Moq.Analyzers.Test;
+
+/// <summary>
+/// Direct unit tests for the internal <c>Moq.CodeFixes.SyntaxGeneratorExtensions</c> argument
+/// validation guards. The throw paths are not exercised by the higher-level code-fix verifier
+/// tests because the real fixers never violate the contracts, so they are validated here with
+/// positive, negative, and boundary cases for every guard in <c>InsertArguments</c> and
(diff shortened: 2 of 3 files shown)


rjmurillo:
Addressed automated review feedback in commit 2e23fd4:

- **Stack trace preservation** (gemini, coderabbit): replaced `throw ex.InnerException` with `ExceptionDispatchInfo.Throw(ex.InnerException)` so an unexpected guard exception keeps its original throw site.
- **Reflection robustness** (Copilot): `GetGuardMethod` now searches public and non-public members and throws `MissingMethodException` immediately if a guard method is renamed, instead of a deferred `NullReferenceException`.
- **Resource churn** (gemini): cache a single `SyntaxGenerator` (`SharedGenerator`) instead of creating and leaking 16 `AdhocWorkspace` instances.

The remaining DeepSource/Codacy notices (CA1303 ResourceManager, magic numbers in assertions, AssemblyVersion) are test-code style noise and are not addressed.

